PR #5, opened 2023-08-30
[feature] Add tracking script

tomjeatt:
Most of these changes are automatic linting. The only actual code change is the plugins config at L36.

### DIFF
--- a/docs/docusaurus.config.js
+++ b/docs/docusaurus.config.js
@@ -1,51 +1,70 @@
 // @ts-check
 // Note: type annotations allow type checking and IDEs autocompletion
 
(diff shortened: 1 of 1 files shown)


nud3l (changes requested):
I think the linting changes are related to the editor/IDE you are using. Could you revert those and then we can eventually set up proper linting rules for the repo?

nud3l (approved):
Changed my mind. Let's merge this and setup linting rules later. The other files also use double quotes over single quotes.

tomjeatt:
> Changed my mind. Let's merge this and setup linting rules later. The other files also use double quotes over single quotes.

No worries—we use the same rules across all of our other UI repos so will be very straightforward to reuse here.

PR #8, opened 2023-09-11
feat: initial bridge draft with lots of todos

sander2:
Lots of todos left

### DIFF
--- a/contracts/bridge/src/bridge.sol
+++ b/contracts/bridge/src/bridge.sol
@@ -1,0 +1,129 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.13;
+
+import {ERC20} from "lib/openzeppelin-contracts/contracts/token/ERC20/ERC20.sol";
+
+contract Bridge is ERC20 {
+    uint256 public number;
+    uint256 public collateralThreshold;
+    uint256 nextRedeemId;
+    mapping(uint256 => RedeemRequest) public redeemRequests; // cant have struct as key, nor tupple
+    mapping(address => uint256) public suppliedCollateral;
+    uint256 totalCollateral;
+
+    struct RedeemRequest {
+        bool open; // actually redundant now that we have accepterAddress
+        uint256 amountZbtc;
+        uint256 amountBtc;
+        address requesterAddress;
+        address accepterAddress;
+        BitcoinAddress bitcoinAddress;
+    }
+    struct BitcoinAddress {
+        uint256 bitcoinAddress; // todo: use the right type
+    }
+    struct TransactionProof {
+        // todo: fields here
+        uint256 dummy;
+    }
+
+    constructor(uint256 threshold) ERC20("Our ZBTC", "ZBTC") {
+        collateralThreshold = threshold;
+    }
+
+    /// lock COL in exchange for zBTC and cCOL
+    function mintZbtc() public payable {
+        uint256 collateral = msg.value; // this is the amount of eth sent to the contract
+        uint256 mintedAmount = colToBtc(collateral) / collateralThreshold;
+        _mint(msg.sender, mintedAmount);
+
+        suppliedCollateral[msg.sender] += collateral;
+        totalCollateral += collateral;
+    }
+
+    /// request zBTC to be redeemed for given amount of BTC.
+    function requestRedeem(
+        uint256 amountZbtc,
+        uint256 amountBtc,
+        BitcoinAddress calldata bitcoinAddress
+    ) public {
+        // lock Zbtc by transfering it to the contract address
+        _transfer(msg.sender, address(this), amountZbtc);
+        require(amountZbtc != 0);
+
+        uint256 id = nextRedeemId++;
+        redeemRequests[id] = RedeemRequest({
+            open: true,
+            amountZbtc: amountZbtc,
+            amountBtc: amountBtc,
+            requesterAddress: msg.sender,
+            accepterAddress: address(0),
+            bitcoinAddress: bitcoinAddress
+        });
+    }
+
+    function acceptRedeem(uint256 id) public {
+        RedeemRequest storage redeemRequest = redeemRequests[id];
+        require(redeemRequest.open);
+
+        // todo: protocol should probably require some sort of collateral deposit here
+
+        redeemRequest.open = false;
+        redeemRequest.accepterAddress = msg.sender;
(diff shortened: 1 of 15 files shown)


contracts/bridge/src/bridge.sol
```diff
@@ -0,0 +1,129 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.13;
+
+import {ERC20} from "lib/openzeppelin-contracts/contracts/token/ERC20/ERC20.sol";
+
+contract Bridge is ERC20 {
```
nud3l, line 6: Instead of making the bridge an ERC20, it is better to separate the two contract deployments where the bridge contract calls into the erc20 contract to mint new tokens and for users to be able to burn their tokens. I think this is achievable with https://docs.openzeppelin.com/contracts/4.x/api/token/erc20#ERC20PresetMinterPauser

contracts/bridge/src/bridge.sol
```diff
@@ -0,0 +1,129 @@
...
+        // todo: fields here
+        uint256 dummy;
+    }
+
+    constructor(uint256 threshold) ERC20("Our ZBTC", "ZBTC") {
+        collateralThreshold = threshold;
+    }
+
+    /// lock COL in exchange for zBTC and cCOL
+    function mintZbtc() public payable {
```
nud3l, line 35: I would call it `mint` instead of hardcoding the name.

contracts/bridge/src/bridge.sol
```diff
@@ -0,0 +1,129 @@
...
+        uint256 collateral = msg.value; // this is the amount of eth sent to the contract
+        uint256 mintedAmount = colToBtc(collateral) / collateralThreshold;
+        _mint(msg.sender, mintedAmount);
+
+        suppliedCollateral[msg.sender] += collateral;
+        totalCollateral += collateral;
+    }
+
+    /// request zBTC to be redeemed for given amount of BTC.
+    function requestRedeem(
```
nud3l, line 45: I'd like to switch over to a notion o calling it swap instead of redeem. So requestRedeem would be requestSwap, ... and the redeemRequests would be orders placed by users.

contracts/bridge/src/bridge.sol
```diff
@@ -0,0 +1,129 @@
...
+        require(redeemRequest.open);
+
+        // todo: protocol should probably require some sort of collateral deposit here
+
+        redeemRequest.open = false;
+        redeemRequest.accepterAddress = msg.sender;
+    }
+
+    // not documented, but presumably required
+    function cancelRedeem(uint256 id) public {}
```
nud3l, line 76: If nobody has accepted the order, user can cancel the order at any time.

contracts/bridge/src/bridge.sol
```diff
@@ -0,0 +1,129 @@
...
+        // burn the zbtc erc20
+        _burn(msg.sender, amountZbtc);
+
+        // transfer eth to caller
+        uint256 collateral = btcToCol(amountZbtc);
+        address payable callerAddress = payable(msg.sender);
+        callerAddress.transfer(collateral);
+    }
+
+    function withdrawFreeCol() public {
```
nud3l, line 111: I think we can just call it `withdraw`

contracts/bridge/src/bridge.sol
sander2: I looked into this today, and it took me some time to understand why it wasn't working for me.. So the problem is that if you deploy the ERC20 contract separately, the bridge contract can not transfer or move user's funds, unless prior to calling the bridge contract, they explicitly call `erc20.approve` to allow the bridge to take control of their funds. Idk if this is a common pattern in the ethereum world, but it seems less user friendly to me

nud3l: Yep, it's common to issue a generic approve in the ETH world so that is fine.

Better use ERC20 due to some wallet incompatibility and also when others integrate with the bridge, better for them to use the ERC20 interface.

contracts/bridge/src/marketplace.sol
```diff
@@ -0,0 +1,68 @@
...
+
+     function withdrawErcErcOrder(uint id) public {
+        Order memory order = ercErcOrders[id];
+        require(order.requesterAddress == msg.sender);
+
+        IERC20(order.offeringToken).safeTransfer(msg.sender, order.offeringAmount);
+
+        delete ercErcOrders[id];
+    }
+}
```
peterslany, line 68: Can you also add `view` function to get all erc-erc orders please? Something like
```solidity
function getActiveErcErcOrders() external view returns (Order[] memory activeErcErcOrders )  {
    activeErcErcOrders = new Order[](nextOrderId);
    for (uint id = 0;id < nextOrderId; id++ ) {
        activeErcErcOrders[id] = ercErcOrders[id];
    }
}
```
We will use this on UI side until we add indexer.

contracts/bridge/src/btc_marketplace.sol
```diff
@@ -0,0 +1,321 @@
...
+        require(accept.requester == msg.sender);
+
+        require(
+            block.timestamp > accept.acceptTime + REQUEST_EXPIRATION_SECONDS
+        );
+
+        // release the locked erc20s
+        IERC20(accept.ercToken).safeTransfer(msg.sender, accept.ercAmount);
+
+        // note: we don't make the accepted amount available for new trades but if we want to,
```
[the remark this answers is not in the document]
sander2: > Shouldn't we delete accepted buy order somewhere here?

Yea, added.

> Also, original order amounts are not updated.

That's what the comment describes, it doesn't make the accepted amount available again. The requester currently has to re-request if they still want to buy, but I think for a POC that's fine

contracts/bridge/src/btc_marketplace.sol
```diff
@@ -0,0 +1,321 @@
...
+    struct BtcBuyOrder {
+        uint256 amountBtc;
+        BitcoinAddress bitcoinAddress;
+        address offeringToken;
+        uint offeringAmount;
+        address requester;
+    }
+
+    struct AcceptedBtcBuyOrder {
+        uint256 amountBtc;
```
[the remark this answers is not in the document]
sander2: As I mentioned in the other comment, I don't think we necessarily have to implement this for the poc

peterslany (approved):

PR #19, opened 2023-10-09
Add docker-compose

gregdhill:
Latest supported version [here](https://github.com/ruimarinho/docker-bitcoin-core) is v23, if we want to upgrade to v25 we will need a different docker build.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,7 +1,7 @@
 version: "3.8"
 services:
     bitcoind:
-        image: "ruimarinho/bitcoin-core:22"
+        image: "ruimarinho/bitcoin-core:23"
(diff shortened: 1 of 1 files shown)


nud3l:
> Latest supported version [here](https://github.com/ruimarinho/docker-bitcoin-core) is v23, if we want to upgrade to v25 we will need a different docker build.

I'd recommend using a newer version right away since ordinals requires core 24 or greater: https://docs.ordinals.com/guides/inscriptions.html?highlight=bitcoin%20core#installing-bitcoin-core

gregdhill:
Will merge this now and @ns212 will submit a PR to update that docker image later

Issue #4, opened 2023-08-30
BOB: Late Middle Ages


nud3l:
# Abstract

Develop a PoC for the BOB stack that allows:

1. Builders to deploy EVM contracts and use a BTC relay to interact with Bitcoin.
2. Users to bridge between a testnet and the rollup (ETH, ERC20) as well as bridge between Bitcoin (BTC) and the rollup via a mock BTC bridge.

# Motivation

The Late Middle Age was a period in Europe preceding the Renaissance. It was characterized by many crisis but saw great progress in the arts and sciences. We chose this name for BOB to reflect the many decisions that will have to be made for the stack and at the same time, the new achievements we will be able to make.

BOB is a large project, and having a PoC will allow us to understand the difficulties in bringing it to the testnet and mainnet readiness.

## Goals

Target completion: End of September 2023

- Builders can spin up a local testnet
- Builders can deploy EVM smart contracts on the local testnet
- Builders can interact with a BTC relay (can still be mocked in parts)
- Users can interact with a local testnet via EVM interfaces (wallets, ethers.js) bridging ETH/ERC20s/BTC
- Completed design for a novel Bitcoin bridge
- Alpha-stage documentation for the rollup design and how to build on it

# Specification

Deliver a proof of concept implementation of the BOB stack rollup and builder platform.

## Organizational

- [x] Create a new GitHub org
- [x] Publish new docs

## Research

- [ ] Specification for BTC bridge
- [x] Decide which rollup stack to use

## Chain

- [x] Spin up OP Stack rollup for testing
- [x] https://github.com/bob-collective/bob/issues/15
- [x] #14

## Smart Contracts

Solidity/EVM

- [x] Mock deployment for BTC bridge: supply collateral, mint synthetic BTC, redeem synthetic BTC, swap synthetic BTC with BTC
- [x] Evaluate existing BTC relay options (Interlay, tBTC, ...) according to maturity, cost, features (taproot, segwit, ...)
- [x] Deploy a BTC relay

## UI

- [x] Create a new UI repo
- [x] Proof of concept
  - Setup ethers.js for account management and runtime interaction
  - WalletConnect 2.0 or MetaMask wallet integration
  - Transfer ERC20 on rollup

gregdhill:
We have chosen to use the production ready **tBTC** (summa / keep-network) relay contracts and supporting libraries to support the initial development of the BOB stack. The contracts are already well-optimized for gas consumption and have been used on mainnet Ethereum for quite some time.

[...]

PR #22, opened 2023-10-09
Add light-relay code from tbtc-v2

gregdhill:
Forked from https://github.com/keep-network/tbtc-v2

### DIFF
--- a/test/LightRelay.t.sol
+++ b/test/LightRelay.t.sol
@@ -1,0 +1,64 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity 0.8.17;
+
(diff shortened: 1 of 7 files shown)


test/LightRelay.t.sol
```diff
@@ -0,0 +1,64 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity 0.8.17;
+
```
sander2, line 3: why did you not use the unmodified one from tbtc?

gregdhill: What do you mean?

gregdhill:
See comment here: https://github.com/bob-collective/bob/issues/4#issuecomment-1754532295

PR #73, opened 2023-10-30
feat: APIs for Ordinals in SDK 

nakul1010:
The PR adds the following things
- [x] Add GET APIs for Ordinals with examples and doc
- [x] Updates Bitcoin core to 24 (required by [ord](https://github.com/ordinals/ord)). 
- [x] Add docker image for ord


sdk/src/ordinals/indexer.ts
```diff
@@ -0,0 +1,325 @@
+/**
+ * Base path for Ordinals regtest Indexer.
+ * @default "http://localhost:3003"
```
sander2, line 3: This isn't the default?

sdk/src/ordinals/indexer.ts
```diff
@@ -0,0 +1,325 @@
...
+    output_value: string,
+    parent: string,
+    previous: string,
+    rune: string,
+    sat: string,
+    satpoint: string
+    timestamp: number
+}
+
+export interface OrdinalsClient {
```
sander2, line 71: All arguments are strings here, that doesn't seem very robust. In some cases, it doesn't make a lot of sense (like for blocknumber), and for others we can probably use stronger (more restrictive) types

sander2 (commented):
Is there a way we can add tests for this?

nakul1010:
> Is there a way we can add tests for this?
Yes once we get a testnet server running, can create a testnet client and use it.

sdk/src/ordinals/indexer.ts
```diff
@@ -0,0 +1,324 @@
...
+/**
+ * @ignore
+ */
+export interface InscriptionUTXO {
+    value: number;
+    script_pubkey: string,
+    address:string,
+    transaction:string,
+    sat_ranges: string,
+    inscriptions: string[];
```
peterslany, line 27: This is supposed to be inscription content, right? In that case I would suggest adding a new type for that, so it does not get confused with inscription ids, something like `type InscriptionContent = string;` which can be later used for `inscriptions: InscriptionContent[]`

sdk/src/ordinals/indexer.ts
```diff
@@ -0,0 +1,324 @@
...
+     * @returns {Promise<string>} A Promise that resolves to a string representing a list of inscriptions.
+     *
+     * @example
+     * ```typescript
+     * const client = new DefaultOrdinalsClient("regtest");
+     * const inscriptions: InscriptionsData = await client.getInscriptions();
+     * console.log("Inscriptions Data:", inscriptions);
+     * ```
+     */
+    getInscriptions(): Promise<InscriptionsData>;
```
peterslany, line 97: Suggestion for a future extension: add filtering, sorting and pagination here, so that it would be possible to filter by inscription number range, creator's address and sort by age.

sdk/src/ordinals/indexer.ts
```diff
@@ -0,0 +1,324 @@
...
+     *
+     * @example
+     * ```typescript
+     * const client = new DefaultOrdinalsClient("regtest");
+     * let block: string = "enter_your_block_number_here";
+     * const inscriptions: InscriptionsData = await client.getInscriptionFromBlock(block);
+     * console.log("Inscriptions Data:", inscriptions);
+     * ```
+     */
+    getInscriptionFromBlock(height: string): Promise<InscriptionsData>;
```
peterslany, line 112:
```suggestion
    getInscriptionsFromBlock(height: string): Promise<InscriptionsData>;
```

sdk/src/ordinals/indexer.ts
```diff
@@ -0,0 +1,324 @@
...
+     *
+     * @example
+     * ```typescript
+     * const client = new DefaultOrdinalsClient("regtest");
+     * let sat: string = "enter_your_sat_number_here";
+     * const inscriptions: InscriptionSat = await client.getInscriptionFromSat(sat);
+     * console.log("Inscriptions Data:", inscriptions);
+     * ```
+     */
+    getInscriptionFromSat(sat: string): Promise<InscriptionSat>;
```
peterslany, line 142:
```suggestion
    getInscriptionsFromSat(sat: string): Promise<InscriptionSat>;
```

sdk/src/ordinals/indexer.ts
```diff
@@ -0,0 +1,324 @@
...
+    address:string,
+    transaction:string,
+    sat_ranges: string,
+    inscriptions: string[];
+}
+
+/**
+ * @ignore
+ */
+export interface InscriptionSat {
```
peterslany, line 33:
```suggestion
export interface Ordinal {
```

sdk/src/ordinals/indexer.ts
nakul1010: Changed

sdk/src/ordinals/indexer.ts
nakul1010: Added a todo in the code, but i don't think it should block this PR

nakul1010:
> @sander2 @peterslany, there are still some remaining todos, but they should be addressed in different issues/PRs

ToDos completed as image upload to docker hub

sdk/test/indexer.test.ts
```diff
@@ -0,0 +1,109 @@
...
+        assert.deepStrictEqual(expectedInscription, inscriptions);
+    });
+
+    it("should get inscriptions from UTXO", async () => {
+        const client = new DefaultOrdinalsClient("testnet");
+        const utxo: string = "d370be1b6bf74677c82226d7a0d65743cbe3846b9216e0ad207a7b03a5230ec3:0";
+        const inscriptions: InscriptionUTXO = await client.getInscriptionFromUTXO(utxo);
+        const expectedInscription: InscriptionUTXO = {
+            value: 1967,
+            script_pubkey: "OP_PUSHNUM_1 OP_PUSHBYTES_32 24ad201633789999cbe4251018e796acb22ec5d1a6f8a1873adc6363e04d7e7d",
```
sander2, line 69: Probably not much this api can do about it here, but it doesn't seem very useful to me to have named OP_X's here, that just makes it harder to parse for code

sander2 (approved):
I don't have the best view on this since I haven't been involved with the API much, but it lgtm from a quick look

peterslany (approved):

sdk/test/indexer.test.ts
nakul1010: removed

bvotteler (approved):
LGTM

PR #79, opened 2023-11-01
docs: add note on using the sdk to inscribe ordinals

gregdhill:


docs/tsconfig.json
```diff
@@ -4,6 +4,10 @@
   "compilerOptions": {
     "baseUrl": "."
   },
-  "include": ["../sdk/src/*"], // Adjust the path as needed
-  "exclude": ["../sdk/src/index.ts"]
-}
+  "include": [
```
tomjeatt, line 7: @gregdhill what are these changes for? There's a build error relating to a missing module in the sdk directory—to fix it we we need to run `yarn` in the `sdk` directory. Not sure we should need to do that though. We can if we need to, but don't want to change the build settings without checking :)

tomjeatt: I'll also sort out access to the Vercel logs—not sure why you don't have that

tomjeatt (approved):

PR #91, opened 2023-11-16
Mt and aa2

sander2:


src/faucet/Erc20Minter.sol
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-pragma solidity 0.8.18;
+pragma solidity ^0.8.13;
```
peterslany, line 2: I was thinking whether we should lock the compiler version or not, I would suggest to do so if we are done with the development, as per the best practices: https://consensys.github.io/smart-contract-best-practices/development-recommendations/solidity-specific/locking-pragmas/

peterslany: Aside from the security implications, it can also help us with verifying the deployed bytecode.

src/swap/Faucet.sol
```diff
@@ -0,0 +1,40 @@
...
+        }
+    }
+
+    function _msgSender() internal view override(Context, ERC2771Recipient) returns (address sender) {
+        sender = ERC2771Recipient._msgSender();
+    }
+
+    function _msgData() internal view override(Context, ERC2771Recipient) returns (bytes calldata) {
+        return ERC2771Recipient._msgData();
+    }
```
peterslany, line 39: Why is it needed to override these functions when they return the same value as if they weren't overrode?

peterslany (approved):
Changes look good, just a couple of non-blocking comments.

src/faucet/Erc20Minter.sol
sander2: Ok yea will keep that in mind, but it was giving me compilation errors so I changed it. We'll see if we can lock version prior to mainnet launch

PR #92, opened 2023-11-16
Update LightRelay.sol

Aryan9592:
Enhances some 'for' loops using `unchecked` block.
Refer [this](https://gist.github.com/grGred/9bab8b9bad0cd42fc23d4e31e7347144#for-loops-improvement)

Also got some typos or grammatical errors fixed.

Thanks!

### DIFF
--- a/src/relay/LightRelay.sol
+++ b/src/relay/LightRelay.sol
@@ -64,7 +64,7 @@ contract LightRelay is Ownable, ILightRelay {
 
     bool public ready;
     // Whether the relay requires the address submitting a retarget to be
-    // authorised in advance by governance.
+    //authorized in advance by governance.
     bool public authorizationRequired;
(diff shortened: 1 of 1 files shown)


nud3l:
Thanks for opening the PR. We are closing this though since these are audited contracts and gas savings on an L2 are still valid but not worth it at this point.

Changing British to American English isn't fixing typos ;)

Aryan9592:
@nud3l For those gas optimizations...I agree that all depends on your codebase (like you said it's on L2 and already audited). Cool!
But in case of changing the typos, I will rather say if `authorisation` -> `authorization` wasn't your concern.. the rest grammatical mistakes are still valid (if you looked them further). Anyways, there's no need to open this PR again just fixing these ones. Main aim was to optimize this which doesn't work here as per this codebase requirement...Thanks!

PR #106, opened 2023-12-06
chore: format forge-rs docs as part of the build process

ns212:


.github/workflows/test.yml
```diff
@@ -33,12 +34,18 @@ jobs:
...
       - name: Run Forge docs
         run: |
           forge doc --out docs/docs/contracts
           python3 docs_reformat_scripts/forge_doc_reformat.py
 
-      - name: Run Forge tests
-        run: |
-          forge test -vvv
-        id: test
+      - name: Commit Forge docs
```
nakul1010, line 47: Didn't commit the docs as expected, the `.gitignore` needs to be updated.

PR #119, opened 2023-12-18
(WIP)chore: test cases for relay.

nakul1010:
The Pr makes the following changes
- Add test cases of light relay.


test/swap/Ord_Marketplace.t.sol
```diff
@@ -305,6 +305,24 @@ contract OrdMarketPlaceTest is OrdMarketplace, Test {
...
+
+        // withdraw sell order
+        this.withdrawOrdinalSellOrder(0);
+
+        // try accepting withdraw order
+        vm.startPrank(bob);
+        token1.approve(address(this), 100);
+        // since order is withdrawn it will try to send tokens to default address
+        vm.expectRevert("Address: call to non-contract");
+        this.acceptOrdinalSellOrder(0, ordinalsInfo[0].requester);
```
nakul1010, line 322: ideally, it should check if the default value is returned from storage and then throw a custom error such as "Order not present". 
But this then does require additional checks in the ordinal contract not sure if we want to add that, as business logic is not getting affected.

test/BitcoinTx.t.sol
```diff
@@ -43,12 +43,125 @@ contract BitcoinTxTest is Test {
...
+    //        );
+    //    }
+
+    // [FAIL. Reason: EvmError: OutOfGas]
+    //    function test_getTxOutputValueWithInvalidInput() public {
+    //        uint64 value = BitcoinTx.getTxOutputValue(
+    //            keccak256(hex"16001493adab0a7a8cb7675db135c9c97e81942025c2c9"),
+    //            hex"16001493adab0a7a8cb7675db135c9c97e81942025c2c9"
+    //        );
+    //    }
```
nakul1010, line 162: during debugging found out that, this happens because the `outputLength`  calculations emit a huge value since the wrong output is given. 
Therefore the assembly calculation [here](https://github.com/bob-collective/bob/blob/master/src/bridge/BitcoinTx.sol#L272-L280) exhausts the gas. 
Not sure want to handle this, as estimated gas will show a huge value if the inputs are wrong.

test/BitcoinTx.t.sol
```diff
@@ -43,12 +43,125 @@ contract BitcoinTxTest is Test {
...
+    }
+
+    // Panic: [FAIL. Reason: panic: array out-of-bounds access (0x32)]
+    //    function test_getTxOutputValueEmptyBytes() public {
+    //        bytes memory emptyBytes = new bytes(0);
+    //        uint64 value = BitcoinTx.getTxOutputValue(
+    //            keccak256(hex"16001493adab0a7a8cb7675db135c9c97e81942025c2c9"),
+    //            emptyBytes
+    //        );
+    //    }
```
nakul1010, line 154: - the `parseVarInt` operation [here](https://github.com/bob-collective/bob/blob/master/src/bridge/BitcoinTx.sol#L221-L222) panics 
- not exactly sure we can handle this

test/BitcoinTx.t.sol
```diff
@@ -43,12 +43,136 @@ contract BitcoinTxTest is Test {
...
+
+    function test_getTxOutputValueWithInvalidInput() public {
+        (bool success,) = address(this).call(
+            abi.encodeWithSignature(
+                "getTxOutputValue(bytes32,bytes memory)",
+                keccak256(hex"16001493adab0a7a8cb7675db135c9c97e81942025c2c9"),
+                hex"16001493adab0a7a8cb7675db135c9c97e81942025c2c9"
+            )
+        );
+        // [FAIL. Reason: EvmError: OutOfGas] (gas: 9223372036854754743)
```
nakul1010, line 171: during debugging found out that, this happens because the `outputLength` calculations emit a huge value since the wrong output is given.
Therefore the assembly calculation [here](https://github.com/bob-collective/bob/blob/master/src/bridge/BitcoinTx.sol#L272-L280) exhausts the gas.
Not sure want to handle this, as the estimated gas field will show a huge value if the inputs are wrong.

foundry.toml
```diff
@@ -3,7 +3,7 @@ src = "src"
 out = "out"
 libs = ["lib"]
 solc = "0.8.17"
-
+viaIR = true
```
gregdhill, line 6: What's this for?

test/BitcoinTx.t.sol
```diff
@@ -43,12 +43,136 @@ contract BitcoinTxTest is Test {
         );
     }
 
-    function test_GetTxOutputValue() public {
+    //
+    // start reverseEndianness
+    //
```
gregdhill, line 48: Remove these and following comments

test/BitcoinTx.t.sol
```diff
@@ -43,12 +43,136 @@ contract BitcoinTxTest is Test {
...
+    }
+
+    function test_ensureTxInputSpendsUtxoReadOverrun() public {
+        bytes memory inputVector = hex"ff";
+        BitcoinTx.UTXO memory utxo;
+        vm.expectRevert("Read overrun during VarInt parsing");
+        BitcoinTx.ensureTxInputSpendsUtxo(inputVector, utxo);
+    }
+
+    function test_ensureTxInputSpendsUtxoEmptyBytes() public {
```
gregdhill, line 98: I think convention is to capitalize the first letter of the test case like so:
```suggestion
    function test_EnsureTxInputSpendsUtxoEmptyBytes() public {
```

foundry.toml
nakul1010: emits out `stack to deep error`
removed added it from the `toml` and explicitly added in ci test command.

gregdhill (approved):
Good job @nakul1010!

PR #166, opened 2024-02-08
fix: compile contract with viaIR flag .

nakul1010:
The `viaIR` flag is currently disabled during the build process, leading to a compiler error. You can check the error [here](https://github.com/bob-collective/bob/actions/runs/7828910688/job/21359772350#step:5:10). It's recommended to enable this flag by default to avoid build errors.

Note: It doesn't add any security vulnerabilities, just a semantic change.

### DIFF
--- a/foundry.toml
+++ b/foundry.toml
@@ -3,7 +3,7 @@ src = "src"
 out = "out"
 libs = ["lib"]
 solc = "0.8.17"
-# `viaIR` flag needs to unabled since `Lightrelay.t.sol` requires it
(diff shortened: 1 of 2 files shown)


foundry.toml
```diff
@@ -3,7 +3,8 @@ src = "src"
 out = "out"
 libs = ["lib"]
 solc = "0.8.17"
-
+# `viaIR` flag needs to unabled since `Lightrelay.t.sol` requires it
```
gregdhill, line 6:
```suggestion
# `viaIR` flag needs to enabled since `Lightrelay.t.sol` requires it
```

PR #183, opened 2024-03-12
Update demo link

derrekcoleman:


derrekcoleman:
Never mind, thought about it more and this isn't necessary (and too small to be a PR, for my own future reference). Sorry!

PR #197, opened 2024-03-28
refactor: remove system state

gregdhill:
Make it easier for downstream consumers, state only makes sense for tbtc


nakul1010 (approved):
Fix CI other than that LGTM

PR #234, opened 2024-05-03
The code for HelloBitcoin.sol has been updated.

ferencdg:


docs/docs/build/getting-started/helloworld.mdx
```diff
@@ -344,7 +341,7 @@ contract HelloBitcoin {
         testLightRelay.setDifficultyFromHeaders(proof.bitcoinHeaders);
 
         // Validate the BTC transaction proof using the relay
-        relay.validateProof(transaction, proof);
+        BitcoinTx.validateProof(relay, 1, transaction, proof);
```
sander2, line 344: The `1` is the number of bitcoin confirmations, right? Would be good to add a comment (also on the one below) explaining that in production you should use a higher number

ferencdg: yes, I will do that, but there are a lot of other problems too ;)

sander2 (approved):

PR #243, opened 2024-05-15
Add CTA links for builders to connect with BD

derrekcoleman:
This adds several callouts on relevant docs pages for builders to contact us. If we want something like this to be even more visible, we could consider a CTA in the header of all pages.

@nud3l, do you think a "Join Fusion" link next to Discord and GitHub in the header is too prominent, potentially leading to spam for Chris?


nud3l (approved):

nud3l:
Thanks!

> @nud3l, do you think a "Join Fusion" link next to Discord and GitHub in the header is too prominent, potentially leading to spam for Chris?

If we get feedback that it's hard to find the fusion link, we could also add it to the header but for now, I think having it in the text is good enough.

PR #267, opened 2024-06-24
refactor: import gateway client and updated wallet logic

gregdhill:
Add example usage

### DIFF
--- a/sdk/examples/gateway.ts
+++ b/sdk/examples/gateway.ts
@@ -40,7 +40,7 @@ async function createTxWithOpReturn(fromAddress: string, toAddress: string, amou
         }
     }
 
-    const unsignedTransaction = await createTransfer(
+    const unsignedTx = await createTransfer(
         'mainnet',
         addressType,
         fromAddress,
@@ -50,19 +50,13 @@ async function createTxWithOpReturn(fromAddress: string, toAddress: string, amou
         opReturn,
     );
 
-    // Determine how many inputs to sign
-    const inputLength = unsignedTransaction.inputsLength;
-    const _inputsToSign = Array.from({ length: inputLength }, (_, i) => i);
+    const psbt = unsignedTx.toPSBT(0);
+    const psbtHex = hex.encode(psbt);
 
-    // Sign all inputs
-    const psbt = unsignedTransaction.toPSBT(0);
-    const _psbtHex = hex.encode(psbt);
-
-    // Sign all inputs with the payment address
-    const signedPsbtHex = ""; // TODO: sign PSBT
+    // TODO: sign PSBT
+    const signedPsbtHex = psbtHex;
 
     const signedTx = SigTx.fromPSBT(bitcoin.Psbt.fromHex(signedPsbtHex).toBuffer());
-
     signedTx.finalize();
 
     return bitcoin.Transaction.fromBuffer(Buffer.from(signedTx.extract()));
--- a/sdk/examples/gateway.ts
+++ b/sdk/examples/gateway.ts
@@ -1,0 +1,69 @@
+import { GatewayApiClient } from "../src/gateway";
+import * as bitcoin from "bitcoinjs-lib";
+import { AddressType, getAddressInfo } from "bitcoin-address-validation";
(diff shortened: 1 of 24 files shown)


sdk/examples/gateway.ts
```diff
@@ -0,0 +1,69 @@
...
+        fromAddress,
+        toAddress,
+        amount,
+        fromPubKey,
+        opReturn,
+    );
+
+    // Determine how many inputs to sign
+    const inputLength = unsignedTransaction.inputsLength;
+    const _inputsToSign = Array.from({ length: inputLength }, (_, i) => i);
```
sander2, line 55: Probably should either explain why this is here, or remove this

sdk/examples/gateway.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    // Sign all inputs
+    const psbt = unsignedTransaction.toPSBT(0);
+    const _psbtHex = hex.encode(psbt);
+
+    // Sign all inputs with the payment address
+    const signedPsbtHex = ""; // TODO: sign PSBT
```
sander2, line 62: I'm not sure what the intended use case is here, if it's to be used in e.g. bots, or if it's to allow people to embed this in their UIs. If the latter, an example showing that would also be nice

sdk/src/utils.ts
```diff
@@ -175,3 +180,46 @@ export function getMerkleProof(block: Block, txHash: string, forWitness?: boolea
         root: merkleAndRoot.root.toString("hex"),
     };
 }
+
+export function estimateTxFee(feeRate: number, numInputs: number = 1) {
```
sander2, line 184: should document the unit of the feeRate (I've think I've seen 3 different units being used in other code before)

sdk/src/utils.ts
```diff
@@ -175,3 +180,46 @@ export function getMerkleProof(block: Block, txHash: string, forWitness?: boolea
...
+    tx.ins.map(tx_input => (tx_input.witness = [Buffer.alloc(33 + 32 + 7, 0), Buffer.alloc(33, 0)]));
+    tx.addOutput(Buffer.alloc(22, 0), 1000); // P2WPKH
+    tx.addOutput(Buffer.alloc(22, 0), 1000); // P2WPKH (change)
+    tx.addOutput(bitcoin.script.compile([bitcoin.opcodes.OP_RETURN, Buffer.alloc(20, 0)]), 0);
+    const vsize = tx.virtualSize();
+    const satoshis = feeRate * vsize;
+    return satoshis;
+}
+
+export function estimateTxFeeConst(feeRate: number, numInputs: number = 1) {
```
sander2, line 199: document how this is different from `estimateTxFee`?

sdk/src/wallet/utxo.ts
```diff
@@ -0,0 +1,159 @@
...
+  toAddress: string,
+  amount: number,
+  publicKey?: string,
+  opReturnData?: string,
+  confirmationTarget: number = 3,
+): Promise<Transaction> {
+  const esploraClient = new DefaultEsploraClient(network);
+
+  // NOTE: esplora only returns the 25 most recent UTXOs
+  // TODO: change this to use the pagination API and return all UTXOs
```
sander2, line 45: Do you want to address this before merging? Or maybe make an issue for it?

sander2 (commented):
Assuming most of the code is copied from the current production code. Mostly lgtm but the exposed functions lack in-code function documentation

sdk/examples/gateway.ts
gregdhill: This is just example code, extracted from the UI to show partners how to integrate

gregdhill: Plan is for @derrekcoleman to use this in a React example instead

sdk/src/utils.ts
gregdhill: Removed

sdk/src/wallet/utxo.ts
gregdhill: Will make an issue

PR #272, opened 2024-06-25
Add checks for broken anchors and duplicate routes

derrekcoleman:
Small tweak to our docs config for future-proofing certain errors at build time


nud3l (approved):

nud3l:
This is a good change, but needs fixing of the errors that are now popping up in the build.

derrekcoleman:
@nud3l as far as I can tell, those are all build errors related to recent SDK changes. If I'm correct in thinking that these weren't caused by the changes in this PR, should we merge this anyway as-is?

PR #273, opened 2024-06-27
feat: full btc relay added

### LINKED ISSUE #285: BOB Full Relay
**Is your feature request related to a problem? Please describe.**
Currently we have a [`LightRelay`](https://github.com/bob-collective/bob/blob/master/src/relay/LightRelay.sol) implementation copied from the tBTC-v2 repository. The problem with this relay is that it is not fully secure under SPV assumptions. I wrote a security analysis [here](https://github.com/bob-collective/bob?tab=readme-ov-file#security) but we recently discovered it is a lot cheaper to spoof a block within 2016 * 2 blocks. This isn't an issue for tBTC because they anyway have it controlled by the signer set and they also track specific UTXOs for the global wallet. 

**Describe the solution you'd like**
We need to build a "Full Relay" based on our previous implementations:

- Solidity: https://github.com/interlay/btc-relay-solidity
- Rust: https://github.com/interlay/interbtc/blob/master/crates/btc-relay/src/lib.rs

See also the old tBTC full relay: https://github.com/summa-tx/relays/blob/30704ab2c443a8a24bdb1228da1e453347f62d3b/solidity/contracts/Relay.sol

**Describe alternatives you've considered**
We could also build some kind of ZK light client using RISC Zero.

**Additional context**
Note the following bugs:
- https://github.com/interlay/interbtc/issues/1186
- https://github.com/interlay/interbtc/issues/1073

ferencdg:
Forked from https://github.com/summa-tx/relays
Changes made:

1. dependency changes
1.1 changed summa-tx/bitcoin-spv to keep-network/bitcoin-spv-sol
1.2 removed SafeMath
2. test changes
 2.1 fixed some tests that were written incorrectly in the summa repo
 2.2 ported Truffle javascript tests to Foundry solidity
 2.3 new tests added
3. solidity compiler version upgraded to 0.8.17
4. OnDemandSPV was gutted and only the verification part was kept

Closes #285


src/relay/FullRelay.sol
```diff
@@ -0,0 +1,461 @@
...
+    /// @return                       True if successfully written, error otherwise
+    function addHeadersWithRetarget(
+        bytes calldata _oldPeriodStartHeader,
+        bytes calldata _oldPeriodEndHeader,
+        bytes calldata _headers
+    ) external returns (bool) {
+        require(
+            isHeaderValidLength(_oldPeriodStartHeader) && isHeaderValidLength(_oldPeriodEndHeader)
+                && isHeaderChainValidLength(_headers),
+            "Bad args. Check header and array byte lengths."
```
gregdhill, line 173: For the errors I think you suggested before @ferencdg about using some custom error type? Do we want to add that for this PR or keep the diff with upstream minimal?

test/fullRelay/FullRelayTestUtils.sol
```diff
@@ -0,0 +1,97 @@
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity ^0.8.17;
+
+import {TestRelay} from "../../src/relay/FullRelayWithVerify.sol";
+
+import {stdJson} from "forge-std/StdJson.sol";
```
gregdhill, line 6: Didn't know about this library before!

gregdhill (approved):
I didn't pick apart the Summa code since I think this was mostly a copy-paste but please point out anything specific you changed or think would be worth a second look. I also think for next steps it could be good to integrate this code with `BitcoinTx.sol` and `WitnessTx.sol` (specifically `validateWitnessProof`) since I think that will let projects do some interesting things with Ordinals in the future.

PR #279, opened 2024-07-04
add info on the fraud proof changes

sander2:
Short version that just links to the optimism page


derrekcoleman:
Outdated because of #280. Closing PR

PR #312, opened 2024-08-13
fix: use scure/btc-signer over bitcoinjs in gateway

nakul1010:


docs/docs/build/bob-sdk/gateway.md
```diff
@@ -140,15 +140,7 @@ async function createTxWithOpReturn(
...
-  const psbtHex = hex.encode(psbt);
-  const signedPsbtHex = psbtHex;
-  const signedTx = SigTx.fromPSBT(
-    bitcoin.Psbt.fromHex(signedPsbtHex).toBuffer()
-  );
-  signedTx.finalize();
-  const tx = bitcoin.Transaction.fromBuffer(Buffer.from(signedTx.extract()));
-
-  return tx;
+  return bitcoin.Transaction.fromPSBT(psbt);
```
nud3l, line 143: I think best to return this as a base64 encoded PSBT since this is what the wallets will expect for signing.

nud3l: This essentially

```ts
return Buffer.from(signedTx.extract());
```

docs/docs/build/bob-sdk/gateway.md
```diff
@@ -93,7 +93,7 @@ We recommend using our [sats-wagmi](./sats-wagmi.md) package to query your user'
 
 ```ts title="/src/utils/gateway.ts"
 import { createTransfer } from "@gobob/bob-sdk";
-import * as bitcoin from "bitcoinjs-lib";
+import * as bitcoin from "@scure/btc-signer";
```
nud3l, line 96: Instead of using `*` import only the types you need.

sdk/examples/gateway.ts
```diff
@@ -51,13 +49,8 @@ async function createTxWithOpReturn(fromAddress: string, toAddress: string, amou
     );
 
     const psbt = unsignedTx.toPSBT(0);
-    const psbtHex = hex.encode(psbt);
 
-    // TODO: sign PSBT
```
gregdhill, line 56: Technically this comment is still valid, the code doesn't sign anything in this example

sdk/examples/gateway.ts
```diff
@@ -51,13 +49,8 @@ async function createTxWithOpReturn(fromAddress: string, toAddress: string, amou
...
-    const signedPsbtHex = psbtHex;
+    const signedTx = Transaction.fromPSBT(psbt);
 
-    const signedTx = SigTx.fromPSBT(bitcoin.Psbt.fromHex(signedPsbtHex).toBuffer());
-    signedTx.finalize();
-
-    return bitcoin.Transaction.fromBuffer(Buffer.from(signedTx.extract()));
+    return Buffer.from(signedTx.extract())
```
gregdhill, line 55: Wondering if it makes more sense to return a string?

gregdhill (changes requested):

sdk/examples/gateway.ts
nakul1010: `Buffer.from(signedTx.extract()).toString('base64')` is this what you expect?

nud3l: We should return what the wallets expect to sign. I think that would be a Base64 encoded buffer. but worthwhile checking what sats-connect, xverse, unisat etc want.

nud3l (approved):

PR #322, opened 2024-08-23
feat: add psbt to start order result

gregdhill:


sdk/src/gateway/client.ts
```diff
@@ -191,11 +193,27 @@ export class GatewayApiClient {
             throw new Error('Invalid OP_RETURN hash');
         }
 
+        let opReturnData = data.opReturnHash;
```
nud3l, line 196: I was more imagining returning a complete psbt including the inputs through this: https://github.com/bob-collective/bob/blob/master/sdk/src/wallet/utxo.ts#L32

Just having the outputs requires the frontend then to find the inputs etc so they can't just submit this to be signed by the wallets.

sdk/src/gateway/client.ts
```diff
@@ -72,6 +73,7 @@ type GatewayCreateOrderResponse = {
 type GatewayStartOrderResult = GatewayCreateOrderResponse & {
     bitcoinAddress: string,
     satoshis: number;
+    psbtBase64: string;
```
nud3l, line 76: If we already want this as bi-directional, we could name this `unsignedTx` or something like that so it would be a valid name for both a BTC as well as an EVM tx. No strong preference right now though. Maybe also cleaner to have `psbtBase64?` and eventually add `rawEvmTx?`or something like that.

gregdhill: Yeah I think cleaner to keep separate

Issue #364, opened 2024-09-24
Esplora fee estimate not found


gregdhill:
Sometimes the fee estimate endpoint will return an incomplete object:

```
curl https://btc-testnet.gobob.xyz/fee-estimates | jq .
```

```
{
  "504": 5.026000000000001,
  "144": 5.026000000000001,
  "1008": 5.026000000000001
}
```

This only happens on testnet, we should allow `getFeeEstimate` to return the nearest value (to the number of confirmations) or default to one.

gregdhill:
Closing for this reason: https://github.com/bob-collective/bob/pull/365#issuecomment-2388815826

PR #365, opened 2024-09-25
chore: add loop to find closest target's fee estimate

slavastartsev:
Closes https://github.com/bob-collective/bob/issues/364


sdk/src/esplora.ts
```diff
@@ -333,7 +333,11 @@ export class EsploraClient {
      */
     async getFeeEstimate(confirmationTarget: number): Promise<number> {
         const response = await this.getJson<Record<number, number>>(`${this.basePath}/fee-estimates`);
-        return response[confirmationTarget];
+        if (response[confirmationTarget]) return response[confirmationTarget];
+        const TARGET_MAX = 1008;
+        for (let target = confirmationTarget + 1; target <= TARGET_MAX; target++) {
+            if (response[target]) return response[target];
+        }
```
gregdhill, line 340: I think a better approach is just to convert the `Record` to an array, sort by the confirmation number, and return the first entry in that array.

gregdhill (changes requested):

sdk/src/esplora.ts
slavastartsev: Speaking of time/memory complexity this approach will take O(n) memory + O(n logn) time
Simple for loop takes just O(n) time

slavastartsev: Fixed. Desc sorting + pop()

gregdhill: You were right with the complexity in your comment before btw ;)

gregdhill: but actually I'm not so sure we want this need to consider, fee rate on testnet should always be 1

gregdhill:
Not sure we should fix it like this actually, I think the method should throw an error if the confirmation target isn't available. We are also trying to get around testnet being broken, if we change anything it should be in the UI or sats-wagmi - maybe hardcode the value to 1 instead in that case as it should be.

PR #376, opened 2024-10-10
feat: add fee output to psbt

gregdhill:
Adding some fees in case frontends want to charge users for onramping

### DIFF
--- a/sdk/src/gateway/client.ts
+++ b/sdk/src/gateway/client.ts
@@ -197,14 +197,16 @@ export class GatewayApiClient {
             typeof params.fromChain === 'string' &&
             params.fromChain.toLowerCase() === Chain.BITCOIN
         ) {
-            psbtBase64 = await createBitcoinPsbt(
-                params.fromUserAddress,
-                gatewayQuote.bitcoinAddress,
-                gatewayQuote.satoshis,
-                params.fromUserPublicKey,
-                data.opReturnHash,
-                gatewayQuote.txProofDifficultyFactor
-            );
+            psbtBase64 = await createBitcoinPsbt({
+                fromAddress: params.fromUserAddress,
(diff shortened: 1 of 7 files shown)


sdk/src/wallet/utxo.ts
```diff
@@ -92,14 +110,17 @@ export async function createBitcoinPsbt(
...
+
     if (opReturnData) {
-        // Strip 0x prefix from opReturn
-        if (opReturnData.startsWith('0x')) {
-            opReturnData = opReturnData.slice(2);
-        }
         outputs.push({
             // OP_RETURN https://github.com/paulmillr/scure-btc-signer/issues/26
-            script: Script.encode(['RETURN', hex.decode(opReturnData)]),
+            script: Script.encode(['RETURN', hex.decode(stripHexPrefix(opReturnData))]),
```
nud3l, line 123: Why is this necessary now and wasn't necessary before?

sdk/src/gateway/client.ts
```diff
@@ -436,17 +439,19 @@ function calculateOpReturnHash(req: GatewayCreateOrderRequest) {
...
         .replace(/ /g, '-')
         .replace(/[^\w-]+/g, '');
 }
+
+function calculateFeeAmount(amount: number, fee?: number) {
+    if (typeof fee === 'undefined' || fee < 1 || fee > 1000) {
+        fee = 0;
+    }
+
+    const feeDecimal = fee / 10000;
```
nud3l, line 454: I'd use big.js here. JS arithmetic with decimals is not great.

sdk/src/wallet/utxo.ts
gregdhill: It does the same thing as the code I removed above

perrystan1489 (approved):

PR #381, opened 2024-10-14
Better fee estimation

nud3l:
Instead of relying on assumptions about the inputs, allow fee estimation to account for all possible inputs.

### DIFF
--- a/sdk/src/wallet/utxo.ts
+++ b/sdk/src/wallet/utxo.ts
@@ -2,7 +2,6 @@ import { Transaction, Script, selectUTXO, TEST_NETWORK, NETWORK, p2wpkh, p2sh } 
 import { hex, base64 } from '@scure/base';
 import { AddressType, getAddressInfo, Network } from 'bitcoin-address-validation';
 import { EsploraClient, UTXO } from '../esplora';
-import { SelectionStrategy } from '@scure/btc-signer/lib/utxo';
 
 export type BitcoinNetworkName = Exclude<Network, 'regtest'>;
 
@@ -70,7 +69,7 @@ export async function createBitcoinPsbt(
     const addressInfo = getAddressInfo(fromAddress);
 
     // TODO: possibly, allow other strategies to be passed to this function
-    const utxoSelectionStrategy: SelectionStrategy = 'default';
+    const utxoSelectionStrategy = 'default';
 
     if (addressInfo.network === 'regtest') {
         throw new Error('Bitcoin regtest not supported');
@@ -336,7 +335,7 @@ export async function estimateTxFee(
     let outputs: Output[] = [];
     // Select all UTXOs if no amount is specified
     // Add outputs to the transaction after all UTXOs are selected to prevent tx creation failures
-    let utxoSelectionStrategy: SelectionStrategy = 'all';
+    let utxoSelectionStrategy = 'all';
     if (amount) {
         // Add the target outputs to the transaction
         // Tx creation might fail if the requested amount is more than the available balance plus fees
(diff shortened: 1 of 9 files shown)


sdk/src/wallet/utxo.ts
```diff
@@ -179,3 +181,122 @@ export function getInputFromUtxoAndTx(
...
+ * strategy to select the UTXOs.
+ */
+export async function estimateTxFee(
+    fromAddress: string,
+    amount?: number,
+    publicKey?: string,
+    opReturnData?: string,
+    confirmationTarget: number = 3,
+    utxoSelectionStrategy: string = 'default'
+): Promise<number> {
```
sander2, line 207: If the feerate changes between the tx estimation and the submission of an actual tx, then we can still get failures. Maybe we should either take feerate as an argument (as the [old function](https://github.com/bob-collective/bob/blob/13e08e27b9f2d770f0655c86ecb149809f4cd425/sdk/src/utils.ts#L198) did, or include it in the return value?

sdk/src/wallet/utxo.ts
```diff
@@ -179,3 +214,165 @@ export function getInputFromUtxoAndTx(
...
+            const hex = await esploraClient.getTransactionHex(utxo.txid);
+            const transaction = Transaction.fromRaw(Buffer.from(hex, 'hex'), { allowUnknownOutputs: true });
+            const input = getInputFromUtxoAndTx(
+                addressInfo.network as BitcoinNetworkName,
+                utxo,
+                transaction,
+                addressInfo.type,
+                publicKey
+            );
+            possibleInputs.push(input);
```
sander2, line 311: Nit: would've been slightly cleaner to return `input` here and do `const possibleInputs = await Promise.all(...)`. Not a blocker by any means :)

sdk/src/wallet/utxo.ts
```diff
@@ -179,3 +214,165 @@ export function getInputFromUtxoAndTx(
...
+    const targetOutputs: Output[] = [
+        {
+            address: toAddress,
+            amount: BigInt(amount ? amount : 0),
+        },
+    ];
+
+    if (opReturnData) {
+        // Strip 0x prefix from opReturn
+        if (opReturnData.startsWith('0x')) {
```
sander2, line 325: If this function is supposed to be used by third parties, then it might be a good idea to make this as robust as possible. Let's do this in case some maniac decides to uppercase the opreturn (making it start with `0X`)?

```suggestion
        if (opReturnData.toLowerCase().startsWith('0x')) {
```

sdk/test/utxo.test.ts
```diff
@@ -14,7 +14,9 @@ describe('UTXO Tests', () => {
             // P2WPKH: https://blockstream.info/address/bc1qar0srrr7xfkvy5l643lydnw9re59gtzzwf5mdq
             'bc1qar0srrr7xfkvy5l643lydnw9re59gtzzwf5mdq',
             // P2SH-P2WPKH: https://blockstream.info/address/3DFVKuT9Ft4rWpysAZ1bHpg55EBy1HVPcr
-            '3DFVKuT9Ft4rWpysAZ1bHpg55EBy1HVPcr',
+            // TODO: Use a real P2SH-P2WPKH address
+            // TODO: Add the pubkey to allow spending from the outputs
+            // '3DFVKuT9Ft4rWpysAZ1bHpg55EBy1HVPcr',
```
sander2, line 19: Should these be addressed before merging?

sdk/src/wallet/utxo.ts
```diff
@@ -179,3 +214,165 @@ export function getInputFromUtxoAndTx(
...
+    confirmationTarget: number = 3
+): Promise<bigint> {
+    const addressInfo = getAddressInfo(fromAddress);
+
+    if (addressInfo.network === 'regtest') {
+        throw new Error('Bitcoin regtest not supported');
+    }
+
+    // We need the public key to generate the redeem and witness script to spend the scripts
+    if (addressInfo.type === (AddressType.p2sh || AddressType.p2wsh)) {
```
sander2, line 270: Just to note: `p2wsh` is mentioned as not being supported in `getInputFromUtxoAndTx`

sdk/test/utxo.test.ts
```diff
@@ -252,10 +259,56 @@ describe('UTXO Tests', () => {
...
         assert.isDefined(transaction);
     });
+
+    it('should estimate the fee for a transaction', async () => {
+        // Addresses where randomly picked from blockstream.info
+        const paymentAddresses = [
+            // P2WPKH: https://blockstream.info/address/bc1qar0srrr7xfkvy5l643lydnw9re59gtzzwf5mdq
+            'bc1qar0srrr7xfkvy5l643lydnw9re59gtzzwf5mdq',
+            // P2SH-P2WPKH: https://blockstream.info/address/3DFVKuT9Ft4rWpysAZ1bHpg55EBy1HVPcr
+            // TODO: As above, add a correct P2SH-P2WPKH address with its pub key
```
sander2, line 276: We should probably add this before merging, right?

sdk/test/utxo.test.ts
danielsimao: yeah, it is a test case that Dom disabled.

sander2 (approved):

PR #392, opened 2024-10-16
Feat/confirmed param balance

danielsimao:


sdk/src/esplora.ts
```diff
@@ -414,7 +422,18 @@ export class EsploraClient {
...
         const chainBalance = response.chain_stats.funded_txo_sum - response.chain_stats.spent_txo_sum;
         const mempoolBalance = response.mempool_stats.funded_txo_sum - response.mempool_stats.spent_txo_sum;
-        return chainBalance + mempoolBalance;
+
+        console.log({
+            chain: chainBalance,
+            mempool: mempoolBalance,
+            total: chainBalance + mempoolBalance,
+        });
+
```
sander2, line 431: do you want to keep this?

sdk/src/esplora.ts
```diff
@@ -388,12 +389,19 @@ export class EsploraClient {
...
+     * @dev This method returns an object containing the confirmed balance, unconfirmed balance, and the total balance.
+     *      The confirmed balance represents the amount that is on-chain, while the mempool balance includes transactions
+     *      that are pending (unconfirmed). The total is the sum of both confirmed and unconfirmed balances.
      *
      * @param {string} address - The Bitcoin address to check.
-     * @returns {Promise<number>} A promise that resolves to the balance.
+     * @returns {Promise<{ chain: number, mempool: number, total: number }>} A promise that resolves to an object containing:
+     *      - `chain`: The confirmed on-chain balance in satoshis.
+     *      - `mempool`: The unconfirmed balance (in mempool) in satoshis.
+     *      - `total`: The total balance, which is the sum of the confirmed and unconfirmed balances.
```
sander2, line 402: I would've preferred `confirmed`, `unconfirmed/pending` and `total` but approving anyway because it doesn't matter :)

sander2 (approved):

sdk/src/esplora.ts
nud3l: I'd convert at least to `debug` level so it's not shown by default in the browser console.

PR #408, opened 2024-11-07
Enable taproot address

slavastartsev:


sdk/src/ordinal-api/index.ts
```diff
@@ -432,8 +432,8 @@ export class OrdinalsClient {
      * console.log("Output:", output);
      * ```
      */
-    async getInscriptionsFromOutPoint(outPoint: OutPoint): Promise<OutputJson> {
-        return await this.getJson<OutputJson>(`${this.basePath}/output/${OutPoint.toString(outPoint)}`);
+    getInscriptionsFromOutPoint(outPoint: OutPoint): Promise<OutputJson> {
```
slavastartsev, line 435: no need to have async/await here. not sure if runtime mitigates that nowadays but adding more generator wrappers (async/await) puts argument callback back to the end of the microtask queue and delays execution

sdk/src/wallet/utxo.ts
```diff
@@ -69,20 +70,28 @@ export async function createBitcoinPsbt(
     const addressInfo = getAddressInfo(fromAddress);
 
     // TODO: possibly, allow other strategies to be passed to this function
-    const utxoSelectionStrategy = 'default';
+    let utxoSelectionStrategy = 'default';
+    if (addressInfo.type === AddressType.p2tr) {
+        utxoSelectionStrategy = 'all';
```
slavastartsev, line 75: with `default` strategy I was not able to create psbt
also tests inside @scure/btc-signer use `all` as a strategy for `selectUTXO`

sdk/src/wallet/utxo.ts
```diff
@@ -69,20 +70,28 @@ export async function createBitcoinPsbt(
...
+        addressInfo.type === AddressType.p2wsh ||
+        addressInfo.type === AddressType.p2tr
+    ) {
         if (!publicKey) {
             throw new Error('Public key is required to spend from the selected address type');
         }
     }
 
     const esploraClient = new EsploraClient(addressInfo.network);
+    const ordinalsClient = new OrdinalsClient(addressInfo.network);
```
slavastartsev, line 94: we can not really check if taproot address has any inscriptions in outputs with `/address/:address` endpoint (it does not exist). [Check](https://ordinals-mainnet.gobob.xyz/address/bc1p9u077kplyqud8qyc404l2t5ud8jerwph6t77cd8xqvnan30lef5sqxfjhm)

sdk/src/wallet/utxo.ts
gregdhill: Why not? Would be good to know why

sdk/src/wallet/utxo.ts
gregdhill: Hmm we really should use that if possible, can you check why that endpoint is not yet enabled? Maybe its enabled on an early version. My concern with the approach below is the many additional HTTP calls we need for each UTXO

sdk/src/wallet/utxo.ts
```diff
@@ -140,7 +153,8 @@ export async function createBitcoinPsbt(
     // https://github.com/paulmillr/scure-btc-signer?tab=readme-ov-file#utxo-selection
     // default = exactBiggest/accumBiggest creates tx with smallest fees, but it breaks
     // big outputs to small ones, which in the end will create a lot of outputs close to dust.
-    const transaction = selectUTXO(possibleInputs, outputs, utxoSelectionStrategy, {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const transaction = selectUTXO(possibleInputs, outputs, utxoSelectionStrategy as any, {
```
gregdhill, line 157: Why 'as any'?

gregdhill (commented):
Some test cases would be good please :)

slavastartsev:
> Some test cases would be good please :)

Tests will be added

sdk/src/wallet/utxo.ts
slavastartsev: I will get into why it's causing in issue. Will update

sdk/src/wallet/utxo.ts
slavastartsev: The latest code of ord [includes](https://github.com/ordinals/ord/blob/master/src/subcommand/server.rs#L166) /address endpoint. 
I do agree that making 1 call to an API is more efficient than `N` calls for each utxo. Will update

sdk/src/wallet/utxo.ts
slavastartsev: The same line is used for `estimateTxFee` -> selectUTXO. Might simply update both to be of type `'all' | 'default'`

sdk/src/wallet/utxo.ts
nud3l: Agree with Greg, we should use the ord endpoint. Were you able to resolve the issue with our endpoint?

sdk/src/wallet/utxo.ts
```diff
@@ -186,6 +194,13 @@ export function getInputFromUtxoAndTx(
         }
         const inner = p2wpkh(Buffer.from(publicKey!, 'hex'), getBtcNetwork(network));
         redeemScript = p2sh(inner);
+    } else if (addressType === AddressType.p2tr) {
+        if (!publicKey) {
```
nud3l, line 198: This check is already done above in line 81.

sdk/src/wallet/utxo.ts
```diff
@@ -267,7 +282,7 @@ export async function estimateTxFee(
     }
 
     // We need the public key to generate the redeem and witness script to spend the scripts
-    if (addressInfo.type === (AddressType.p2sh || AddressType.p2wsh)) {
+    if (addressInfo.type === AddressType.p2sh || addressInfo.type === AddressType.p2tr) {
```
nud3l, line 285: As above, why remove p2wsh from this check?

sdk/src/wallet/utxo.ts
slavastartsev: Changes are in my stash. At the moment /address endpoint is still not available. Nikolai said that he updated ord service with the latest changes from upstream.

sdk/src/wallet/utxo.ts
```diff
@@ -69,20 +71,21 @@ export async function createBitcoinPsbt(
...
-    const utxoSelectionStrategy = 'default';
+    const utxoSelectionStrategy: SelectionStrategy = 'default';
 
     if (addressInfo.network === 'regtest') {
         throw new Error('Bitcoin regtest not supported');
     }
 
     // We need the public key to generate the redeem and witness script to spend the scripts
-    if (addressInfo.type === (AddressType.p2sh || AddressType.p2wsh)) {
+    if (addressInfo.type === AddressType.p2sh || addressInfo.type === AddressType.p2tr) {
```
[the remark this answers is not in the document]
slavastartsev: The answer is [here](https://github.com/bob-collective/bob/pull/408#discussion_r1832385126)

sdk/src/wallet/utxo.ts
slavastartsev: Will remove

sdk/src/wallet/utxo.ts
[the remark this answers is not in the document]
slavastartsev: `if (addressInfo.type === (AddressType.p2sh || AddressType.p2wsh)) {` more detailed reply

this `if` statement only checks if `addressInfo.type === AddressType.p2sh`. The reason is that `(AddressType.p2sh || AddressType.p2wsh)` this expression always resolves as `AddressType.p2sh` -- result of `or` operator is the first non-falsy value. I also notices that in tests for utxo file `publicKey` was not provided in case if address type is `p2wsh`.

sdk/src/wallet/utxo.ts
slavastartsev: > Agree with Greg, we should use the ord endpoint. Were you able to resolve the issue with our endpoint?

Changes added:
- request to `/address/:address` endpoint
- update interfaces to conform v0.21.2 ordinals api

sdk/src/wallet/utxo.ts
```diff
@@ -113,7 +122,9 @@ export async function createBitcoinPsbt(
                 addressInfo.type,
                 publicKey
             );
-            possibleInputs.push(input);
+            // to support taproot addresses we want to exclude outputs which contain inscriptions
+            if (!inscriptionsSet.has(InscriptionId.toString({ txid: utxo.txid, index: utxo.vout })))
+                possibleInputs.push(input);
```
sander2, line 127: Are you sure that this works? I think the inscriptions [refer to the txids in which they were revealed](https://docs.ordinals.com/inscriptions.html#inscription-ids), rather than the txid they currently reside in:

> The part in front of the i is the transaction ID (txid) of the reveal transaction.

If that's the case then this logic doesn't work.

I think we need to use [`/outputs/<ADDRESS>`](https://docs.ordinals.com/guides/api.html) which has an option to "return UTXOs not containing inscriptions or runes"

sdk/src/wallet/utxo.ts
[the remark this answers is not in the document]
slavastartsev: > Ok so `AddressType.p2wsh` was never actually checked for. The question is, though, _should_ it be checked for?

Will add this check. It should not break anything.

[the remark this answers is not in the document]
slavastartsev: But based on this comment it looks like p2wsh have yet to be supported https://github.com/bob-collective/bob/blob/154b7339a31cb771b91f576fc8cf9587e2bf25fd/sdk/src/wallet/utxo.ts#L194-L195

sdk/src/wallet/utxo.ts
sander2: It's available [upstream](https://github.com/ordinals/ord/blob/master/src/subcommand/server.rs#L224). Why are we using a fork?

slavastartsev: > Are you sure that this works? I think the inscriptions [refer to the txids in which they were revealed](https://docs.ordinals.com/inscriptions.html#inscription-ids), rather than the txid they currently reside in:
> 
> > The part in front of the i is the transaction ID (txid) of the reveal transaction.
> 
> If that's the case then this logic doesn't work.
> 
> I think we need to use [`/outputs/<ADDRESS>`](https://docs.ordinals.com/guides/api.html) which has an option to "return UTXOs not containing inscriptions or runes"

updated

sdk/src/ordinal-api/index.ts
```diff
@@ -432,8 +472,30 @@ export class OrdinalsClient {
...
+     * @example
+     * ```typescript
+     * const client = new OrdinalsClient("regtest");
+     * const address: string = "enter_address_here";
+     * const type: 'cardinal' | 'inscribed' | 'runic' = "enter_type_here";
+     * const output = await client.getUTXOFromAddress(address, type?);
+     * console.log("Output:", output);
+     * ```
+     */
+    getInscriptionsFromAddress(address: string, type?: 'cardinal' | 'inscribed' | 'runic'): Promise<OutputJson[]> {
```
sander2, line 494: This naming is a bit odd, especially when called with `type = cardinal`, since it does exactly the opposite of what the name suggests: it then returns utxos that _don't_ have inscriptions.

Maybe something like `getOutputsByType` and possibly make `type` non-optional (add `any` option).

sdk/src/ordinal-api/index.ts
```diff
@@ -432,8 +472,30 @@ export class OrdinalsClient {
...
+     * @param {String} address - The Bitcoin address to check.
+     * @param {('cardinal' | 'inscribed' | 'runic')} [type] - Optional type of UTXOs to be returned. If omitted returns all types.
+     * @returns {Promise<OutputJson>} A Promise that resolves to the inscription data.
+     *
+     * @example
+     * ```typescript
+     * const client = new OrdinalsClient("regtest");
+     * const address: string = "enter_address_here";
+     * const type: 'cardinal' | 'inscribed' | 'runic' = "enter_type_here";
+     * const output = await client.getUTXOFromAddress(address, type?);
```
sander2, line 490: update naming

sdk/src/wallet/utxo.ts
```diff
@@ -113,7 +126,8 @@ export async function createBitcoinPsbt(
                 addressInfo.type,
                 publicKey
             );
-            possibleInputs.push(input);
+            // to support taproot addresses we want to exclude outputs which contain inscriptions
+            if (!outpointsSet.has(OutPoint.toString(utxo))) possibleInputs.push(input);
```
sander2, line 130: with `cardinal` argument:
> return UTXOs **not** containing inscriptions or runes

Hence the condition should be flipped here (i.e. remove the `!`)

sdk/src/wallet/utxo.ts
```diff
@@ -69,36 +70,48 @@ export async function createBitcoinPsbt(
...
 
     if (feeRate) {
-        confirmedUtxos = await esploraClient.getAddressUtxos(fromAddress);
+        [confirmedUtxos, outputsS] = await Promise.all([
+            esploraClient.getAddressUtxos(fromAddress),
+            ordinalsClient.getInscriptionsFromAddress(fromAddress, 'cardinal'),
+        ]);
     } else {
-        [confirmedUtxos, feeRate] = await Promise.all([
+        [confirmedUtxos, feeRate, outputsS] = await Promise.all([
```
sander2, line 102: Do we still need the `esploraClient.getAddressUtxos` if we already have the data from the ordinals api?

Also, another thing we should check before merging, is how the ord api handles utxos in the mempool. Ideally we want the behavior to be the same as the current esplora api

sdk/src/wallet/utxo.ts
```diff
@@ -69,36 +70,48 @@ export async function createBitcoinPsbt(
...
         if (!publicKey) {
             throw new Error('Public key is required to spend from the selected address type');
         }
     }
 
     const esploraClient = new EsploraClient(addressInfo.network);
+    const ordinalsClient = new OrdinalsClient(addressInfo.network);
 
     let confirmedUtxos: UTXO[] = [];
+    let outputsS: OutputJson[] = [];
```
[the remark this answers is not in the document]
slavastartsev: sorry, will fix that. forgot to stage a file with updates

sdk/src/ordinal-api/index.ts
slavastartsev: updated

sdk/src/wallet/utxo.ts
slavastartsev: We can not 100% sure about that if we do not simulate in mainnet, right? Currently checking their docs and tests to find a confirmation to that

sander2: I would say the opposite - testing this on mainnet does not make us 100% sure. It's edge cases that I'm worried about. If we only have 1 data source, we don't need to worry about the edge case where one of the api updates before the other one does. E.g. if esplora updates more quickly than ord, then if an ordinal is sent to the address, esplora will find the utxo but ord does not have it marked as being an ordinal yet, which could lead to us spending it if we're not careful.

That's not to say that we can't keep 2 data sources, just that we need to be careful. I think the current approach _should_ be fine - we only use utxos that are reported as spendable by both esplora and ord

sdk/src/wallet/utxo.ts
nud3l, line 1: I'd add some debug info regarding the ordinals here around things that might go wrong/edge cases like the possible utxo discrepancy between electrs and ord: https://github.com/bob-collective/bob/blob/master/sdk/src/wallet/utxo.ts#L358 and https://github.com/bob-collective/bob/blob/master/sdk/src/wallet/utxo.ts#L158.

Since the code lives in three different repos, easier to debug if we have most of the info already in the UI and don't need to patch the SDK version.

sdk/src/wallet/utxo.ts
```diff
@@ -69,20 +70,28 @@ export async function createBitcoinPsbt(
...
         throw new Error('Bitcoin regtest not supported');
     }
 
     // We need the public key to generate the redeem and witness script to spend the scripts
-    if (addressInfo.type === (AddressType.p2sh || AddressType.p2wsh)) {
+    if (
+        addressInfo.type === AddressType.p2sh ||
+        addressInfo.type === AddressType.p2wsh ||
+        addressInfo.type === AddressType.p2tr
+    ) {
```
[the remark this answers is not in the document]
nud3l: Oh, you are right. We still need the public key to spend a p2wsh though since we can use the public key to generate the witness script that we need to spend the p2wsh (assuming this is a p2wsh-p2pkh or similar). If it was a multisig or other things under the hood, we'd need more info in the witness script generation.

sdk/src/wallet/utxo.ts
```diff
@@ -69,36 +70,48 @@ export async function createBitcoinPsbt(
...
         if (!publicKey) {
             throw new Error('Public key is required to spend from the selected address type');
         }
     }
 
     const esploraClient = new EsploraClient(addressInfo.network);
+    const ordinalsClient = new OrdinalsClient(addressInfo.network);
 
     let confirmedUtxos: UTXO[] = [];
+    let outputsFromAddress: OutputJson[] = [];
```
nud3l, line 94: If the address isn't a p2tr, you'd technically don't need to query the ordinals client.

sdk/src/wallet/utxo.ts
```diff
@@ -69,36 +70,48 @@ export async function createBitcoinPsbt(
...
+        [confirmedUtxos, outputsFromAddress] = await Promise.all([
+            esploraClient.getAddressUtxos(fromAddress),
+            ordinalsClient.getOutputsFromAddress(fromAddress, 'cardinal'),
+        ]);
     } else {
-        [confirmedUtxos, feeRate] = await Promise.all([
+        [confirmedUtxos, feeRate, outputsFromAddress] = await Promise.all([
             esploraClient.getAddressUtxos(fromAddress),
             esploraClient.getFeeEstimate(confirmationTarget),
+            ordinalsClient.getOutputsFromAddress(fromAddress, 'cardinal'),
```
nud3l, line 105: Suggest to leave a comment here like "cardinal = return UTXOs not containing inscriptions or runes"

sdk/src/wallet/utxo.ts
slavastartsev: will put conditional check before ord api call 👍

sdk/src/wallet/utxo.ts
slavastartsev: I've added to the logs both `confirmedUtxos` and `outputsFromAddress`. Because these lists of outputs could get quite big in size we can not really know upfront what kinds of discrepancies users might be looking for. Fine to work with logged lis in the browser, not sure about nodejs libraries through since these 2 combined could exceed stout max length/size

nud3l (approved):

sdk/src/wallet/utxo.ts
```diff
@@ -279,33 +310,51 @@ export async function estimateTxFee(
...
+        ]);
     } else {
-        [confirmedUtxos, feeRate] = await Promise.all([
+        [confirmedUtxos, feeRate, outputsFromAddress] = await Promise.all([
             esploraClient.getAddressUtxos(fromAddress),
             esploraClient.getFeeEstimate(confirmationTarget),
+            // cardinal = return UTXOs not containing inscriptions or runes
+            addressInfo.type === AddressType.p2tr ? ordinalsClient.getOutputsFromAddress(fromAddress, 'cardinal') : [],
         ]);
     }
```
gregdhill, line 331: Suggestion but we could reduce the code a bit here:

```ts
[confirmedUtxos, feeRate, outputsFromAddress] = await Promise.all([
    esploraClient.getAddressUtxos(fromAddress),
    feeRate !== undefined ? Promise.resolve(feeRate) : esploraClient.getFeeEstimate(confirmationTarget),
    // cardinal = return UTXOs not containing inscriptions or runes
    addressInfo.type === AddressType.p2tr ? ordinalsClient.getOutputsFromAddress(fromAddress, 'cardinal') : [],
]);
```

sdk/src/wallet/utxo.ts
```diff
@@ -279,33 +310,51 @@ export async function estimateTxFee(
     // TODO: allow submitting the UTXOs, fee estimate and confirmed transactions
     // to avoid fetching them again.
     const esploraClient = new EsploraClient(addressInfo.network);
+    const ordinalsClient = new OrdinalsClient(addressInfo.network);
 
     let confirmedUtxos: UTXO[] = [];
+    let outputsFromAddress: OutputJson[] = [];
```
gregdhill, line 316: Would leave a comment here that this contains UTXOs which *do not* contain inscriptions. I assume that's right from reading the code below anyway

sdk/src/wallet/utxo.ts
```diff
@@ -279,33 +310,51 @@ export async function estimateTxFee(
...
                 addressInfo.type,
                 publicKey
             );
+
+            // to support taproot addresses we want to exclude outputs which contain inscriptions
+            if (addressInfo.type === AddressType.p2tr) {
+                if (outpointsSet.has(OutPoint.toString(utxo))) possibleInputs.push(input);
+            } else {
+                possibleInputs.push(input);
+            }
```
gregdhill, line 357: Minor nit / suggested change:
```suggestion
            if (addressInfo.type === AddressType.p2tr && outpointsSet.has(OutPoint.toString(utxo))) {
                possibleInputs.push(input);
            } else {
                possibleInputs.push(input);
            }
```

sdk/src/wallet/utxo.ts
```diff
@@ -279,33 +310,51 @@ export async function estimateTxFee(
...
-            return getInputFromUtxoAndTx(
+            const input = getInputFromUtxoAndTx(
                 addressInfo.network as BitcoinNetworkName,
                 utxo,
                 transaction,
                 addressInfo.type,
                 publicKey
             );
+
+            // to support taproot addresses we want to exclude outputs which contain inscriptions
```
gregdhill, line 352: Btw do you know if the Ord API will index the mempool txs appropriately? Seems to me there could be a race condition if esplora returns a UTXO from the mempool and Ord has not yet updated that - cc: @sander2

sdk/src/wallet/utxo.ts
slavastartsev: This is basically a different condition.
Let's say address type is `p2tr` then `input` will be treated as `possible input` regardless whether it's a part `outpointsSet` or not.

gregdhill: Ah no you're right nevermind

sdk/src/wallet/utxo.ts
slavastartsev: comment added

sdk/src/wallet/utxo.ts
slavastartsev: updated

sdk/src/wallet/utxo.ts
slavastartsev: The answers are [1](https://github.com/bob-collective/bob/pull/408#discussion_r1841929462) and [2](https://github.com/bob-collective/sats-wagmi/pull/125#discussion_r1842863707)

sdk/test/utxo.test.ts
```diff
@@ -1,9 +1,19 @@
...
+import { OrdinalsClient, OutPoint } from '../src/ordinal-api';
+
+vi.mock(import('@scure/btc-signer'), async (importOriginal) => {
+    const actual = await importOriginal();
+
+    return {
+        ...actual,
+        selectUTXO: vi.fn(actual.selectUTXO),
+    };
+});
```
gregdhill, line 16: What is the purpose of this?

sdk/test/utxo.test.ts
```diff
@@ -323,4 +333,38 @@ describe('UTXO Tests', () => {
...
+
+        // cardinal = return UTXOs not containing inscriptions or runes
+        const cardinalOutputs = await ordinalsClient.getOutputsFromAddress(paymentAddress, 'cardinal');
+
+        const cardinalOutputsSet = new Set(cardinalOutputs.map((output) => output.outpoint));
+
+        const maxSpendableBalance = cardinalOutputs.reduce((acc, output) => acc + output.value, 0);
+
+        // spend 90% of max spendable amount
+        await createBitcoinPsbt(paymentAddress, paymentAddress, Math.floor(maxSpendableBalance * 0.9), pubkey);
```
gregdhill: I think we should check this fails, check the balance of the address including ordinals, try to create the UTXO and check it throws an error that the account has insufficient balance. I think this test is not really doing anything useful right now since we're checking only the cardinal balance.

gregdhill (changes requested):

sdk/test/utxo.test.ts
slavastartsev: it allows us to check what arguments it was called with.
- vi.fn - mock/spi function
- actual.selectUTXO - original function implementation

so when it will be called it will run the same code + we can spy on arguments/errors/return values

sdk/test/utxo.test.ts
slavastartsev: This test is checking that when we create psbt none of the possible inputs contains `inscriptions` -- @sander2 wanted that to be tested.

I've added a test that checks if it throws an errors if we try to spends max balance including ordinals. I think this new test is more general and works (throws) with all types of addresses.

gregdhill: Hmm actually yeah you are right that this test is still valid

sander2 (approved):

PR #419, opened 2024-11-25
Add balance calculation logic

slavastartsev:
Also includes updates to 
- estimateTxFee
- createBitcoinPsbt

⚠️ Inputs that contain ordinals or runes no longer will be used in tx fee estimations or could be spent from (any, not just taproot) address

### DIFF
--- a/sdk/test/utxo.test.ts
+++ b/sdk/test/utxo.test.ts
@@ -2,9 +2,10 @@ import { vi, describe, it, assert, Mock, expect } from 'vitest';
 import { AddressType, getAddressInfo, Network } from 'bitcoin-address-validation';
 import { Address, NETWORK, OutScript, Script, Transaction, p2sh, p2wpkh, selectUTXO } from '@scure/btc-signer';
 import { hex, base64 } from '@scure/base';
-import { createBitcoinPsbt, getInputFromUtxoAndTx, estimateTxFee, Input } from '../src/wallet/utxo';
+import { createBitcoinPsbt, getInputFromUtxoAndTx, estimateTxFee, Input, getBalance } from '../src/wallet/utxo';
 import { TransactionOutput } from '@scure/btc-signer/psbt';
 import { OrdinalsClient, OutPoint } from '../src/ordinal-api';
+import { beforeEach } from 'node:test';
 
 vi.mock(import('@scure/btc-signer'), async (importOriginal) => {
     const actual = await importOriginal();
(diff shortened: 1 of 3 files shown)


gregdhill (commented):
Would be good to have a test that checks the logic i.e. returns a smaller number if the address then holds ordinals. Maybe just alter the mocks to not return one of the UTXOs or something?

sdk/test/utxo.test.ts
```diff
@@ -392,4 +415,57 @@ describe('UTXO Tests', () => {
...
+                const outpoint = OutPoint.toString(output);
+
+                return { outpoint };
+            })
+        );
+
+        const balanceData = await getBalance(taprootAddress);
+
+        expect(balanceData.total).toBeLessThanOrEqual(total);
+        expect(balanceData.confirmed).toBeLessThanOrEqual(confirmed);
```
gregdhill, line 468: Should it not be:
```suggestion
        expect(balanceData.confirmed).toBeLessThanOrEqual(confirmed/2);
```
etc.

slavastartsev: it's not that we are taking half of the confirmed amount but half of the confirmed UTXOs. theoretically total amount of that half could be more than half of the confirmed amount.

example (4 UTXOs and their values):
`1 1 3 4`
then if we take right half of that list`3 + 4 -> total 7` will be greater than `1 + 1 + 3 + 4) / 2 -> total 4.5`

gregdhill: Ah right yes you are correct, but what I mean is that validating `toBeLessThanOrEqual` doesn't really tell us much right? Let's say for some reason the code we are unit testing is broken and allows spending all UTXOs including those containing runes or ordinals, then the test will still pass.

slavastartsev: notes for that test:
- fetching all confirmed UTXOs from explora client -- some of the might contain ordinals
- mocking response of ord api so that left half of that list contains ordinals (api will return only right path) -- (example above is incorrect, should be `4 3 1 1`, `4 3` - contain inscriptions, `1 1` - do not contain inscriptions)
- executing `getBalance` fn.

so if we are assuming that some of these UTXOs contain inscription then total value of `all` UTXOs should be less than or equal (if there're 0 UTXOs) to the total amount of UTXOs (cardinal) that ord api will respond with.

> Would be good to have a test that checks the logic i.e. returns a smaller number if the address then holds ordinals. Maybe just alter the mocks to not return one of the UTXOs or something?

slavastartsev: > Ah right yes you are correct, but what I mean is that validating `toBeLessThanOrEqual` doesn't really tell us much right? Let's say for some reason the code we are unit testing is broken and allows spending all UTXOs including those containing runes or ordinals, then the test will still pass.

So let's say ord api with `cardinal` query params returns the exact same UTXOs as not containing inscription (which is incorrect) then yes -- this behaviour is incorrect.

This test is not testing ord implementation of `/outputs/:address?type=cardinal` -- we are assuming it works properly all the time. Such behaviour ideally should be caught by ord unit tests.

gregdhill (approved):

PR #433, opened 2024-12-12
Better get balance

slavastartsev:
- for each unconfirmed tx, check there are no inscriptions: https://github.com/bob-collective/bob/blob/743d0204cf74b1c7bb7316599369dd34de2b0af5/sdk/src/inscription.ts#L188
- for each input to that tx do step (1) if its also unconfirmed else check via the Ord API as we do now

Explanation: build a tree structure 
- for each output returned by esplora becomes root `TreeNode`
- inputs to that tx are it's children
- leaf nodes either confirmed outputs or outputs containing ordinals (flagged with `cardinal: false`). 

If output has all leaf nodes with `cardinal: true` -- confirmed (not containing ordinals) then output amount could be spent.


sdk/src/wallet/utxo.ts
```diff
@@ -90,28 +141,32 @@ export async function createBitcoinPsbt(
...
+    if (utxos.length === 0) {
         throw new Error('No confirmed UTXOs');
     }
 
-    const outpointsSet = new Set(outputsFromAddress.map((output) => output.outpoint));
+    const cardinalOutputsSet = new Set(cardinalOutputs.map((output) => output.outpoint));
+
+    const rootUtxoNodes = utxos.map((utxo) => new TreeNode<OutputNodeData>({ ...utxo, cardinal: true }));
+
+    await processNodes(rootUtxoNodes, esploraClient);
```
slavastartsev, line 163: 161: store tree roots in a list

163: populate tree

sdk/test/utxo.test.ts
```diff
@@ -435,36 +441,4 @@ describe('UTXO Tests', () => {
         assert(zeroBalance.unconfirmed === 0n, 'If no address specified unconfirmed must be 0');
         assert(zeroBalance.total === 0n, 'If no address specified total must be 0');
     });
-
-    it('returns smalled amount if address holds ordinals', async () => {
```
gregdhill, line 439: Why did you remove this test case?

gregdhill (commented):
I also see similar logic changes in at least 2-3 parts of `utxo.ts` is there anyway to share logic?

sdk/test/utxo.test.ts
slavastartsev: this test will fail because in this one we were checking that half of the outputs were ordinals.
because now there's a follow-up check with inputs to that output containing ordinals it's no longer valid.
I'll restore it if I find a way to mock many more apis i.e. `getTransactionHex` etc

slavastartsev:
> I also see similar logic changes in at least 2-3 parts of `utxo.ts` is there anyway to share logic?

Already done. All shared functions `checkUtxoNode`, `processNodes` and class `TreeNode` extracted.

In each function we only need to create `rootUtxoNodes: TreeNodes[]` that were not confirmed by `ord` and populate them with data.
After that we can check that node and either consider output as possible input or increase spendable amount.

sdk/src/wallet/utxo.ts
```diff
@@ -17,6 +18,57 @@ export const getBtcNetwork = (name: BitcoinNetworkName) => {
...
+    }
+}
+
+const processNodes = async (rootNodes: (TreeNode<OutputNodeData> | null)[], esploraClient: EsploraClient) => {
+    const queue = Array.from(rootNodes);
+
+    while (queue.length > 0) {
+        const childNode = queue.shift();
+
+        if (childNode === null) continue;
```
slavastartsev, line 37: we need to replace `TreeNode`s with `null`s for confirmed outputs to keep indexes aligned with `utxo` list

sdk/test/utxo.test.ts
slavastartsev: upd: updated test

with this change we can assume that `confirmed` balance value will be less that `confirmed` total amount from ord service but `total`s must be equal

sdk/src/wallet/utxo.ts
```diff
@@ -17,6 +18,55 @@ export const getBtcNetwork = (name: BitcoinNetworkName) => {
...
+const processNodes = async (rootNodes: TreeNode<OutputNodeData>[], esploraClient: EsploraClient) => {
+    const queue = Array.from(rootNodes);
+
+    while (queue.length > 0) {
+        const childNode = queue.shift();
+
+        const txInscriptions = await getTxInscriptions(esploraClient, childNode.val.txid);
+
+        if (txInscriptions.length === 0) {
+            const transaction = await esploraClient.getTransaction(childNode.val.txid);
```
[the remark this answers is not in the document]
gregdhill: See the highlighted section: both `getTxInscriptions` and `getTransaction` will call Esplora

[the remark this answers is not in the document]
slavastartsev: > See the highlighted section: both `getTxInscriptions` and `getTransaction` will call Esplora

right
`getTxInscriptions` retrieves `tx hex`
`getTransaction` retrieves tx details
they both provide different data

If you want to say we need to wrap these calls in `Promise.all` it wont really change much -- still 2 calls, no batching
technically we can prefetch `esploraClient.getTransaction` before knowing if `tx` contains any inscriptions but if tx does not contain inscription then we are making one extra backend call which could have been avoided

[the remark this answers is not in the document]
slavastartsev:
```typescript
const txHex = await esploraClient.getTransactionHex(childNode.val.txid);
const tx = bitcoin.Transaction.fromHex(txHex);
```

with reconstructed `tx` it is not possible to check if it was `confirmed` or not. 
without that flag we will run `while` loop until coinbase tx reached

sdk/src/wallet/utxo.ts
```diff
@@ -17,6 +18,69 @@ export const getBtcNetwork = (name: BitcoinNetworkName) => {
...
+        this.val = val;
+        this.children = children;
+    }
+}
+
+const createUtxoNodes = (utxos: UTXO[], cardinalOutputsSet: Set<string>) =>
+    utxos.reduce(
+        (acc, utxo) => {
+            if (!cardinalOutputsSet.has(OutPoint.toString(utxo)))
+                acc.push(new TreeNode<OutputNodeData>({ ...utxo, cardinal: true }));
```
gregdhill, line 35: I would default to `cardinal: false` and then set to `true` if we are sure these aren't ordinals.

sdk/src/wallet/utxo.ts
```diff
@@ -17,6 +18,69 @@ export const getBtcNetwork = (name: BitcoinNetworkName) => {
...
+
+        if (childNode === null) continue;
+
+        const txInscriptions = await getTxInscriptions(esploraClient, childNode.val.txid);
+
+        if (txInscriptions.length === 0) {
+            const transaction = await esploraClient.getTransaction(childNode.val.txid);
+
+            // if not confirmed check inputs for current utxo
+            if (!transaction.status.confirmed) {
```
gregdhill, line 57: Since we don't ever check Ord again the mempool tx might be spending an ordinal, you only check the confirmation here.

gregdhill (changes requested):

sdk/src/wallet/utxo.ts
```diff
@@ -17,6 +17,76 @@ export const getBtcNetwork = (name: BitcoinNetworkName) => {
...
+
+        if (childNode === null) continue;
+
+        const transaction = await esploraClient.getTransaction(childNode.val.txid);
+
+        if (transaction.status.confirmed) {
+            // if confirmed check if it contains ordinals
+            childNode.val.cardinal = cardinalOutputsSet.has(OutPoint.toString(childNode.val));
+        } else {
+            const response = await ordinalsClient.getInscriptionsFromOutPoint(childNode.val);
```
sander2, line 62: What happens if the utxo is unconfirmed? Would the next line panic?

sdk/src/wallet/utxo.ts
```diff
@@ -17,6 +17,76 @@ export const getBtcNetwork = (name: BitcoinNetworkName) => {
...
+
+            if (response.inscriptions.length === 0) {
```
sander2, line 64: Should also check for runes, right?

sander2 (commented):
Would it be possible to add some tests?

sdk/src/wallet/utxo.ts
slavastartsev: eagle eye

sdk/src/wallet/utxo.ts
slavastartsev: it won't
```bash
curl -s -H "Accept: application/json" "https://ordinals-mainnet.gobob.xyz/output/9cd84b0d89695c732f42906e755894319f6afa1aa69fef0b9a0bc5974eed8ef9:2" 
```

sdk/src/wallet/utxo.ts
slavastartsev: fixed

slavastartsev:
> Would it be possible to add some tests?

tests will be added once we agree on the final solution

sdk/src/wallet/utxo.ts
```diff
@@ -17,6 +17,81 @@ export const getBtcNetwork = (name: BitcoinNetworkName) => {
...
+
+        return new TreeNode<OutputNodeData>({
+            ...utxo,
+            cardinal: isCardinal(output),
+            indexed: output.indexed,
+        });
+    });
+};
+
+const processNodes = async (
```
gregdhill, line 48: One suggestion I had which I don't think I see in the tests is to have a bit more coverage. Basically what I would like to see is some data driven unit tests, where we can provide say two arrays (utxos from esplora and utxos from ord) and then run check the core logic. This is better I think to test all the different edge cases, confirmed vs unconfirmed vs inscription in the same block.

gregdhill (changes requested):
Looks good but I would like to see some cleanup and more tests before we merge this.

sdk/src/wallet/utxo.ts
```diff
@@ -17,6 +17,81 @@ export const getBtcNetwork = (name: BitcoinNetworkName) => {
...
+        const childNode = queue.shift();
+
+        if (childNode === null) continue;
+
+        const transaction = await esploraClient.getTransaction(childNode.val.txid);
+
+        if (childNode.val.indexed && transaction.status.confirmed) {
+            // if confirmed check if it contains ordinals
+            childNode.val.cardinal = cardinalOutputsSet.has(OutPoint.toString(childNode.val));
+        } else if (!childNode.val.indexed || childNode.val.cardinal) {
```
sander2, line 66: Wouldn't this work?

```suggestion
        if (!childNode.val.indexed) {
```

slavastartsev: it will
but we also need to check if node does not contain `inscriptions` or `runes` (cardinal = true) to process it's inputs

if we replace this check with just `!childNode.val.indexed` then this will process all txs until coinbase tx reached

slavastartsev: I think we need to check both esplora and ord response. also we need to check indexed and confirmed tx itself if it does not contain runes and inscriptions

sdk/src/wallet/utxo.ts
slavastartsev: Done
Extracted into `processUtxos`

sdk/src/wallet/utxo.ts
```diff
@@ -90,28 +165,32 @@ export async function createBitcoinPsbt(
...
+    if (utxos.length === 0) {
         throw new Error('No confirmed UTXOs');
     }
 
-    const outpointsSet = new Set(outputsFromAddress.map((output) => output.outpoint));
+    const cardinalOutputsSet = new Set(cardinalOutputs.map((output) => output.outpoint));
+
+    const rootUtxoNodes = await createUtxoNodes(utxos, cardinalOutputsSet, ordinalsClient);
+
+    await processNodes(rootUtxoNodes, cardinalOutputsSet, esploraClient, ordinalsClient);
```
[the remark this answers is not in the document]
slavastartsev: Done

sdk/src/wallet/utxo.ts
```diff
@@ -90,28 +187,30 @@ export async function createBitcoinPsbt(
...
+    const cardinalOutputsSet = new Set(cardinalOutputs.map((output) => output.outpoint));
+
+    const allowedList = await processUtxos(utxos, cardinalOutputsSet, esploraClient, ordinalsClient);
 
     // To construct the spending transaction and estimate the fee, we need the transactions for the UTXOs
     const possibleInputs: Input[] = [];
 
     await Promise.all(
-        confirmedUtxos.map(async (utxo) => {
+        utxos.map(async (utxo, index) => {
```
gregdhill, line 213: Why are we still doing this? The point of now having `allowedList` is to only use that, same below

sdk/test/utxo.test.ts
```diff
@@ -454,17 +444,222 @@ describe('UTXO Tests', () => {
...
+
+            // mark as unconfirmed -> continue building tree for `vin`s
+            if (tx === utxos[1].txid || tx === utxos[2].txid) {
+                result.status.confirmed = false;
+            }
+
+            return result;
+        });
+
+        const allowedList = await processUtxos(utxos, cardinalOutputsSet, esploraClient, ordinalsClient);
```
gregdhill, line 661: We're not mocking the ordinals service right? If the ordinals in this test case move will that break the tests?

slavastartsev: > We're not mocking the ordinals service right? If the ordinals in this test case move will that break the tests?

Not mocking ordinals service.
Tests should not fail because we are checking txs which are immutable once mined.

sdk/src/wallet/utxo.ts
slavastartsev: We need `allowedList` also for [getBalance](https://github.com/bob-collective/bob/blob/4f6c956c52ec59ff2ab43b754375eacbac3023df/sdk/src/wallet/utxo.ts#L536-L547) function. IMO easier to keep this logic outside.

sdk/src/wallet/utxo.ts
```diff
@@ -90,28 +187,30 @@ export async function createBitcoinPsbt(
...
         ordinalsClient.getOutputsFromAddress(fromAddress, 'cardinal'),
     ]);
 
-    if (confirmedUtxos.length === 0) {
+    if (utxos.length === 0) {
         throw new Error('No confirmed UTXOs');
     }
 
-    const outpointsSet = new Set(outputsFromAddress.map((output) => output.outpoint));
+    const cardinalOutputsSet = new Set(cardinalOutputs.map((output) => output.outpoint));
```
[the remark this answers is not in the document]
slavastartsev: extracted as much as I could
open for naming suggestions

sdk/src/wallet/utxo.ts
```diff
@@ -17,6 +17,150 @@ export const getBtcNetwork = (name: BitcoinNetworkName) => {
...
+
+    const [utxos, cardinalOutputs] = await Promise.all([
+        esploraClient.getAddressUtxos(address),
+        // cardinal = return UTXOs not containing inscriptions or runes
+        ordinalsClient.getOutputsFromAddress(address, 'cardinal'),
+    ]);
+
+    // if (utxos.length === 0) {
+    //     throw new Error('No confirmed UTXOs');
+    // }
```
slavastartsev, line 130: had to comment this out
I think it still necessary to keep it

sdk/src/wallet/utxo.ts
```diff
@@ -294,46 +357,13 @@ export async function estimateTxFee(
     // Use the from address as the toAddress for the fee estimate
     const toAddress = fromAddress;
 
-    // TODO: allow submitting the UTXOs, fee estimate and confirmed transactions
-    // to avoid fetching them again.
```
gregdhill, line 298: Keep todo note

sdk/src/wallet/utxo.ts
```diff
@@ -416,25 +445,13 @@ export async function getBalance(address?: string) {
...
 
-        return acc;
-    }, 0);
+    const total = safeUtxos.reduce((acc, utxo) => acc + utxo.value, 0);
 
-    const confirmed = outputs.reduce((acc, output) => {
-        if (cardinalOutputsSet.has(OutPoint.toString(output)) && output.confirmed) {
-            return acc + output.value;
+    const confirmed = safeUtxos.reduce((acc, utxo) => {
+        if (utxo.confirmed) {
```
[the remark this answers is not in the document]
slavastartsev: this code is calculating `confirmed` balance.

`total` is on line [451](https://github.com/bob-collective/bob/blob/d71f1cd1222d81e9b81d6b14edb374c508771951/sdk/src/wallet/utxo.ts#L451-L452)

sdk/src/wallet/utxo.ts
```diff
@@ -17,6 +17,102 @@ export const getBtcNetwork = (name: BitcoinNetworkName) => {
...
+
+const isCardinalTx = async (
+    txid: string,
+    cardinalOutputsSet: Set<string>,
+    esploraClient: EsploraClient,
+    ordinalsClient: OrdinalsClient,
+    limit: number = 3
+): Promise<boolean> => {
+    if (limit === 0) return false;
+    const transaction = await esploraClient.getTransaction(txid);
```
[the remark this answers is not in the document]
slavastartsev: not sure if it's possible. from the current schema there're no endpoints which might returns us `inputs` for `txid`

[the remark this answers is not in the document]
gregdhill: Ah I think it returns the hex encoded tx which can be decoded to a transaction object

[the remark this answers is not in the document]
slavastartsev: unfortunately it's of another interface `Transaction`. coincident that name is the same

gregdhill (approved):

PR #462, opened 2025-01-23
Check mempool tx for rune

slavastartsev:


sdk/src/runes.ts
```diff
@@ -0,0 +1,36 @@
+import * as bitcoin from 'bitcoinjs-lib';
+import { EsploraClient } from './esplora';
+
+class Rune {
+    data: string;
```
gregdhill, line 5: I would make this more specific, would be more useful imo to parse the rune into type / amount etc.

gregdhill: Also because there may be uses of `OP_RETURN OP_13` I suspect that aren't related to runes

sdk/src/runes.ts
```diff
@@ -0,0 +1,36 @@
...
+}
+
+export function parseRune(tx: bitcoin.Transaction): Rune | undefined {
+    for (const out of tx.outs) {
+        try {
+            const asm = bitcoin.script.toASM(out.script);
+
+            if (asm.startsWith('OP_RETURN OP_13')) {
+                const runeData = asm.split(' ')[2];
+                return new Rune(runeData);
```
gregdhill, line 23: Can there be multiple Runes in a tx?

gregdhill (changes requested):
Can you also add some unit tests once you add the Rune parsing?

sdk/src/runes.ts
slavastartsev: I'll double check that
According to the docs they probably can
https://docs.ordinals.com/runes.html
> A transaction output may hold balances of any number of runes.

sdk/src/wallet/utxo.ts
```diff
@@ -68,9 +74,13 @@ export const findSafeUtxos = async (
             // the utxo is confirmed and a known cardinal
             if (cardinalOutputsSet.has(OutPoint.toString(utxo))) return true;
 
-            const inscriptions = await getTxInscriptions(esploraClient, utxo.txid);
+            const txHex = await esploraClient.getTransactionHex(utxo.txid);
+            const tx = bitcoin.Transaction.fromHex(txHex);
+
+            const inscriptions = parseInscriptions(tx);
+            const rune = parseRune(tx);
```
sander2, line 81: fwiw I think this duplication could've been avoided by refactoring `isCardinalTx`. If the check is done at the beginning of the function rather than inside the loop. Which I think would've been cleaner anyway. Not a blocker though

slavastartsev: Good catch 🙌
Updated

sdk/src/runes.ts
slavastartsev: There could be only 1 per tx

> A transaction may have at most one runestone.

slavastartsev: There could be only 1 per tx
from the docs ▼
> A transaction may have at most one runestone.

Although there could be [inscription + rune](https://ordiscan.com/tx/e3206469ee558cd3ea562d7b63e33353f2156b26017470e9674aa098d0efdece) in the same tx if etched 

if transferred [tx may contain many ](https://ordiscan.com/tx/1206c05e2a1dfb31dca892f173a85fcc03dfac159a8c772e0d76e3b33a2d4f08)

sdk/src/runes.ts
slavastartsev: Added `runestone-js` package which is based on `ord@0.17.1` implementation
notes below

sdk/src/runes.ts
```diff
@@ -0,0 +1,20 @@
...
+import { EsploraClient } from './esplora';
+// This implementation is based on ord 0.17.1, *use this as a reference only
+import { Runestone } from 'runestone-js';
+
+export function parseRunestone(tx: bitcoin.Transaction): Runestone | undefined {
+    for (const out of tx.outs) {
+        if (out.script[0] === bitcoin.opcodes.OP_RETURN && out.script[1] === bitcoin.opcodes.OP_13) {
+            // skip OP_RETURN OP_13 OP_PUSHBYTES_31
+            const pureOutputBuffer = out.script.subarray(3);
+            return Runestone.dechiper(pureOutputBuffer);
```
slavastartsev, line 11: Added `runestone-js`. As opposed to `@ordjs/runestone` written in typescript.
Contains typo `Runestone.dechiper` -> `Runestone.decipher`

gregdhill (approved):

PR #479, opened 2025-02-04
fix: utxo test cases to resolve CI issues

nakul1010:


sdk/test/utxo.test.ts
```diff
@@ -230,10 +230,18 @@ describe('UTXO Tests', () => {
         const inner = p2wpkh(Buffer.from(publicKey, 'hex'), NETWORK);
         const redeemScript = p2sh(inner);
 
+        // Generate unique txid values for each UTXO
+        const txids = [
+            '0000000000000000000000000000000000000000000000000000000000000001',
+            '0000000000000000000000000000000000000000000000000000000000000002',
+            '0000000000000000000000000000000000000000000000000000000000000003',
+            '0000000000000000000000000000000000000000000000000000000000000004',
+        ];
```
gregdhill, line 239: Why?

sdk/test/utxo.test.ts
```diff
@@ -242,7 +250,7 @@ describe('UTXO Tests', () => {
                     },
                 },
                 {
-                    txid: Buffer.alloc(32, 0).toString('hex'),
+                    txid: txids[1],
```
gregdhill, line 253:
```suggestion
                    txid: Buffer.alloc(32, 1).toString('hex'), ```

gregdhill (changes requested):

sdk/test/utxo.test.ts
slavastartsev: I think we need to do something like this or tests will throw 
```Error: Estimator: same input passed multiple times: 0000000000000000000000000000000000000000000000000000000000000000:0```

gregdhill (approved):

PR #549, opened 2025-04-02
feat: support proof verification using Full Relay in WitnessTx lib

Orland0x:


contracts/src/utils/WitnessTx.sol
```diff
@@ -106,13 +107,28 @@ library WitnessTx {
...
     ) internal view returns (bytes32 wTxHash) {
         wTxHash = validateWitnessProof(txInfo, proof);
+
+        // Checks that the header chain is valid with respect to the difficulty stored in the light relay
         BitcoinTx.evaluateProofDifficulty(relay, txProofDifficultyFactor, proof.coinbaseProof.bitcoinHeaders);
     }
+
+    function validateWitnessProof(
+        IFullRelayWithVerify relay,
+        uint256 txProofDifficultyFactor,
```
gregdhill, line 124: Technically for the full relay "difficulty factor" doesn't make so much sense, but won't block the PR on this.

gregdhill (approved):

PR #558, opened 2025-04-07
feat(sdk): add APY and intencives to strategies

mastilver:


sdk/src/gateway/types.ts
```diff
@@ -72,6 +72,18 @@ export interface Token {
...
+        name: string;
+        logo: string;
+        apy: number;
+    }[];
+    incentives: {
+        id: 'spice' | 'solvXP' | 'babylonPoints' | 'avalonPoints' | 'vedaPoints' | 'segmentPoints' | 'bedrockDiamonds';
+        name: string;
+        logo: string;
+        amountPerBTCPerDay: number | null;
+    }[];
```
gregdhill, line 86:
```suggestion
    baseApy: {
        name: string;
        value: number;
    };
    rewardApy: { .. },
    points: {
        id: string;
        name: string;
    }[];
```

mastilver: `points` make sense

I would change the rest to

```
    baseApy: number;
    rewardApy: number;
    rewardTokens: Token[]
```

reasoning is:
- baseApy: I don't know what I would put in the name
- rewardApy: DL does not give us a way to differentiate in which form the yield is coming from
- rewardTokens: DL return us rewardTokens as string[], I think we could easily transform it into `Token`s

gregdhill: Sounds good to me, let's align with what Defillama gives us!

sdk/src/gateway/strategy.ts
```diff
@@ -1,8 +1,198 @@
...
+            },
+        ],
+    ],
+    [
+        // Solv (xSolvBTC)
+        '0x0bef2a8b771e37763c1ce02a88f404c6b2573843',
+        [
+            {
+                id: 'solv',
+                name: 'Solv XP',
```
gregdhill, line 90: Does it make sense to move the name out of this mapping and having a lookup, since we will have the same name referenced multiple times?

sdk/src/gateway/strategy.ts
```diff
@@ -1,8 +1,198 @@
...
+        'veda',
+        [
+            {
+                id: 'veda',
+                name: 'Veda Points',
+            },
+        ],
+    ],
+    [
+        'none',
```
gregdhill, line 24: Should be 'bob'

gregdhill (approved):

PR #605, opened 2025-05-13
feat: gateway swap strategy using icecreamswap router

Orland0x:


gregdhill:
Closing as we will instead use the calldata strategies

PR #617, opened 2025-05-22
fix: overload getQuote method

nakul1010:


nakul1010:
`any` type as param can't be used

slavastartsev:
> `any` type as param can't be used

Merged types for function implementation and it worked

PR #639, opened 2025-06-10
feat(relayer): add lib.rs

Nuhvi:
This PR add the following changes necessary for integration tests for contracts using the full relay:
1. Adds a `lib.rs` to `bob-relayer` crate.
2. Change `Relayer::new()` signature to be less dependent on generics of alloy provider.
3. Add a helper function `Relayer::deploy_contract()`.
4. Use an alternative to `goldsky` hardcoded api when using a local BOB rpc url.
5. Remove the requirement for a minimum difficulty in the FullRelay contract.
6. Update bindings.


relayer/src/relayer.rs
```diff
@@ -36,30 +57,93 @@ impl RelayHeader {
...
         let relayer_blockhash = self.contract.getBestKnownDigest().call().await?;
         let relayed_height: u32 =
             self.contract.findHeight(relayer_blockhash).call().await?.try_into().unwrap();
 
         Ok(relayed_height)
     }
 
-    async fn has_relayed(&self, blockhash: BlockHash) -> Result<bool> {
+    /// Returns the block hash of the most recent relayed block.
+    pub async fn relayed_blockhash(&self) -> Result<BlockHash> {
```
sander2:
```suggestion
    pub async fn best_blockhash(&self) -> Result<BlockHash> {
```

relayer/src/relayer.rs
```diff
@@ -86,15 +170,14 @@ impl<
                 return Ok(height);
             }
 
-            println!("Found fork: {actual_hash} at height {height}");
+            tracing::debug!("Found fork: {actual_hash} at height {height}");
```
sander2: I think this can be `info`, this is useful to know

relayer/src/relayer.rs
```diff
@@ -110,21 +193,27 @@ impl<
 
             latest_height += headers.len() as u32;
 
-            if headers.is_empty() {
-                // we are up to date, sleep for a bit
-                tokio::time::sleep(Duration::from_secs(15)).await;
+            self.push_headers(headers).await?;
+
+            tokio::time::sleep(if self.is_regtest() {
+                Duration::from_millis(10)
```
sander2: I think 10 ms is excessive, it would create quite some load without significant benefit. I would suggest something like 100-250 ms

crates/bindings/src/lib.rs
```diff
@@ -1,4 +1,4 @@
-#![allow(unused_imports, clippy::all, rustdoc::all)]
+#![allow(unused_imports, clippy::all, dead_code, rustdoc::all)]
```
sander2, line 1: this will get overwritten when regenerating bindings, but hopefully by then a version of alloy with [this fix](https://github.com/alloy-rs/core/pull/976) has been released so that this is no longer needed

contracts/src/relay/FullRelay.sol
```diff
@@ -49,11 +49,6 @@ contract FullRelay is IFullRelay {
         require(isHeaderValidLength(_genesisHeader), "Bad genesis block");
         bytes32 _genesisDigest = _genesisHeader.hash256();
 
-        require(
-            _periodStart & bytes32(0x0000000000000000000000000000000000000000000000000000000000ffffff) == bytes32(0),
-            "Period start hash does not have work. Hint: wrong byte order?"
-        );
-
```
gregdhill, line 56: Why did we remove this?

sander2: because otherwise we can't deploy on regtest. It's anyway a completely trusted function so it doesn't really matter

PR #680, opened 2025-07-10
feat: gateway SDK v4 changes

nakul1010:
### 📝 Description
* Retrieve V4 order details during the `getOnrampQuote` call.
* Pass the same order details when invoking `startOrder`.

## Summary by CodeRabbit

* **New Features**
  * Enhanced quote and order creation processes to include detailed order information.
  * Added support for specifying an ETH amount to receive when requesting a quote.
  * Updated API endpoints to versioned paths for improved compatibility.

* **Bug Fixes**
  * Improved handling and formatting of order details for greater accuracy and consistency.
  * Added validation to ensure quote responses contain required order details.

* **Tests**
  * Updated test coverage to validate new order details structure and versioned API endpoints.
  * Removed a redundant test case for combined onramp and offramp orders.
  * Disabled specific Bitcoin addresses in fee estimation tests for accuracy.


sdk/src/gateway/client.ts
```diff
@@ -187,10 +196,19 @@ export class GatewayApiClient {
...
-        const quote: GatewayQuote = await response.json();
+        const jsonResponse = await response.json();
+
+        if (!jsonResponse.orderDetails) {
+            throw new Error('Missing orderDetails in quote response');
+        }
+
+        const quote: GatewayQuote = {
+            ...jsonResponse,
+            orderDetails: convertOrderDetailsRawToOrderDetails(jsonResponse.orderDetails),
```
nakul1010, line 207: This is still open for discussion. My assumption, I would prefer that SDK users see bigint values rather than hexadecimal representations given by API.

sdk/src/gateway/utils.ts
```diff
@@ -59,3 +59,41 @@ export function viemClient(chain: Chain) {
...
+function parseU256(value: string | null): bigint | null {
+    if (value === null) return null;
+    return BigInt(value);
+}
+
+export function convertOrderDetailsRawToOrderDetails(order: OrderDetailsRaw): OrderDetails {
+    return {
+        version: order.version,
+        data: {
+            ethAmountToReceive: parseU256(order.data.ethAmountToReceive)!,
```
slavastartsev: `!` at the end of `parseU256` means that the final value is not `null`
while `parseU256` is allowed to return `null` is value is `null`

not sure if it's good approach, might need another solution

sdk/src/gateway/client.ts
```diff
@@ -176,9 +184,10 @@ export class GatewayApiClient {
         const outputTokenAddress = getTokenAddress(this.chainId, toToken);
         const strategyAddress = params.strategyAddress?.startsWith('0x') ? params.strategyAddress : undefined;
 
-        const url = new URL(`${this.baseUrl}/quote/${outputTokenAddress}`);
+        const url = new URL(`${this.baseUrl}/quote/v4/${outputTokenAddress}`);
```
slavastartsev: if api versioning is added it's typically done in a way `/v{version}/%base_path%/%endpoint%`
not sure if adding version in the middle of resource path is good approach of handling this

nakul1010: I need to fix in backend as well, keep the PR open till then

sdk/src/gateway/utils.ts
nakul1010: Check it now `parseU256` can't return a NULL value

slavastartsev: if it can not return null better to remove `!` marks from function calls
```suggestion
            ethAmountToReceive: parseU256(order.data.ethAmountToReceive),
```
also applies to the lines below

PR #682, opened 2025-07-14
feat: add `orderDetails` to `OnrampOrderResponse`

nakul1010:


sdk/src/gateway/client.ts
```diff
@@ -823,9 +823,14 @@ export class GatewayApiClient {
                 }
                 return txStatus.confirmed ? latestHeight - txStatus.block_height! + 1 : 0;
             };
+
+            const orderDetails: OrderDetails | null =
+                order.orderDetails != null ? convertOrderDetailsRawToOrderDetails(order.orderDetails) : null;
```
slavastartsev: `order.orderDetails != null` will always evaluate to `true` because json can not hold `null` value. (`orders` is a parsed json structure)

nakul1010: Is this the right way to go then: 
```ts
const orderDetails: OrderDetails | null =
    order.orderDetails != null && order.orderDetails !== ""
        ? convertOrderDetailsRawToOrderDetails(order.orderDetails)
        : null;

```

slavastartsev: > Is this the right way to go then:
> 
> ```ts
> const orderDetails: OrderDetails | null =
>     order.orderDetails != null && order.orderDetails !== ""
>         ? convertOrderDetailsRawToOrderDetails(order.orderDetails)
>         : null;
> ```

I believe it must be `undefined` not `null`
so more like
 ```ts
 const orderDetails =
     order.orderDetails
         ? convertOrderDetailsRawToOrderDetails(order.orderDetails)
         : undefined;
```

because `order.orderDetails` is of type `OrderDetails | undefined`

slavastartsev (approved):

PR #712, opened 2025-08-18
fix: rename `maxSatsToSwapToEth` to `satsToSwapToEth`

nakul1010:
## Summary by CodeRabbit

* **Refactor**
  * Renamed the API field maxSatsToSwapToEth to satsToSwapToEth across order details types and related conversion logic. This is a breaking change; update integrations accordingly.
* **Documentation**
  * Updated the field description to “Amount of satoshis to be swapped to ETH.”
* **Tests**
  * Updated test mocks to use satsToSwapToEth instead of maxSatsToSwapToEth.

### DIFF
--- a/sdk/src/gateway/types.ts
+++ b/sdk/src/gateway/types.ts
@@ -212,7 +212,7 @@ export type OrderDetailsRaw = {
     version: string;
     data: {
         ethAmountToReceive: string;
-        maxSatsToSwapToEth: number;
+        satsToSwapToEth: number;
         ethTransferGasLimit: string;
         strategyGasLimit: string;
         totalUserGasLimit: string;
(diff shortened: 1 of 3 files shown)


sdk/src/gateway/types.ts
```diff
@@ -229,8 +229,8 @@ export type OrderDetails = {
     data: {
         /** @description The amount of ETH (in wei) that the user will receive */
         ethAmountToReceive: bigint;
```
Orland0x, line 231: also need to remove this. Or is that being done on some other PR?

gregdhill: I think we're still estimating in the backend, but is a good point that if we're doing that we don't need `ethAmountToReceive` and `satsToSwapToEth`

nakul1010: We need to keep `ethAmountToReceive`, as it represents the gas refill value in **wei**.
In the v3 quote API, users can specify the amount in **sats**, so api convert it to wei and returns back.

nakul1010: To be descriptive: 
* **V3 API** → user specifies gas refill in **sats** → API converts to **wei** → returns the converted value.
* **V4 API** → user specifies gas refill directly in **wei** → API simply returns the same value.

The resulting gas refill value (in wei) is then passed to the `create_order` API, where it is required for fee recalculation and verification.

PR #765, opened 2025-09-17
feat: pass user btc address to offramp quote for compliance check

nakul1010:
## Summary by CodeRabbit

- **New Features**
  - SDK now lets integrators include a BTC destination address when requesting an offramp quote. An optional field can be provided to route payouts to a specific user BTC address for more precise payout handling.

- **Chores**
  - Package version bumped to 4.2.10.

### DIFF
--- a/sdk/src/gateway/client.ts
+++ b/sdk/src/gateway/client.ts
@@ -360,7 +360,7 @@ export class GatewayApiClient {
         token: Address,
         amountInToken: bigint,
         userAddress: Address,
-        toUserAddress?: string | null
+        toUserAddress?: string
     ): Promise<OfframpQuote> {
         const queryParams = new URLSearchParams({
             amountInWrappedToken: amountInToken.toString(),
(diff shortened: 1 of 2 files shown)


sdk/src/gateway/client.ts
```diff
@@ -288,7 +288,8 @@ export class GatewayApiClient {
         const quote = await this.fetchOfframpQuote(
             tokenAddress,
             BigInt(params.amount || 0),
-            params.fromUserAddress as Address
+            params.fromUserAddress as Address,
+            params.toUserAddress
```
slavastartsev, line 292: `toUserAddress` is of type `string`
you can update `fetchOfframpQuote` function signature to receive `toUserAddress: string`

slavastartsev: also you can cast it `as Address`

slavastartsev (approved):

sdk/src/gateway/client.ts
nakul1010: `fetchOfframpQuote` is also used in for `getBumpFeeRequirement` [here](https://github.com/bob-collective/bob/blob/6aac766d29a874f7227ea38f0f18e7a9e3015167/sdk/src/gateway/client.ts#L605). 

The `toUserAddress` is optional, hence I've defined it optional

nakul1010: > also you can cast it as Address

`toUserAddress ` is BTC address and not EVM address

PR #770, opened 2025-09-18
feat(gateway-v4): expose layerzero

mastilver:
Add two new function `isLayerZero` and `getLayerZeroDetails`.
- `getLayerZeroDetails` expose eid and if possible the chain id.


sdk/src/gateway/client.ts
```diff
@@ -1289,4 +1325,45 @@ export class GatewayApiClient {
...
+                    'https://metadata.layerzero-api.com/v1/metadata/deployments',
+                    {
+                        headers: {
+                            'Content-Type': 'application/json',
+                            Accept: 'application/json',
+                        },
+                    },
+                    `Failed to get layerzero deployments`
+                );
+
```
nakul1010: There's a layer zero client if you could use that would be ideal. 
check `layerzero.ts`

sdk/src/gateway/layerzero.ts
```diff
@@ -24,97 +25,6 @@ type SendParam = {
...
-
-    private async getJson<T>(url: string): Promise<T> {
-        const response = await fetch(url);
-        if (!response.ok) {
-            throw new Error(response.statusText);
-        }
-        return (await response.json()) as Promise<T>;
-    }
-}
-
```
mastilver, line 120: to prevent cycle imports

sdk/src/gateway/layerzero-client.ts
```diff
@@ -0,0 +1,102 @@
...
+            return {
+                name: chainKey,
+                eid: chainInfo?.eid,
+                oftAddress: deployment.address,
+                nativeChainId: chainInfo?.nativeChainId,
+            };
+        });
+    }
+
+    async getChainId(eid: string): Promise<number | null> {
```
mastilver: added `getChainId`

sdk/src/gateway/client.ts
```diff
@@ -124,6 +125,8 @@ export class GatewayApiClient {
     private strategy: StrategyClient;
     private isSignet: boolean = false;
 
+    private layerzero = new LayerZeroClient();
```
gregdhill: Ah no @mastilver I meant L0 should be completely separate from standard Gateway client

sdk/src/gateway/client.ts
```diff
@@ -1047,6 +1050,9 @@ export class GatewayApiClient {
                 ? convertOrderDetailsRawToOrderDetails(order.orderDetails)
                 : undefined;
 
+            // eslint-disable-next-line @typescript-eslint/no-this-alias
+            const that = this;
```
gregdhill: Why do we need this? Not sure I like it

sdk/src/gateway/client.ts
```diff
@@ -1078,6 +1084,23 @@ export class GatewayApiClient {
...
+                        return null;
+                    }
+
+                    const chainId = await that.layerzero.getChainId(eid.toString());
+
+                    return {
+                        dstEid: eid,
+                        chainId,
+                    };
+                },
```
gregdhill: Honestly I would get rid of this and just return the dst eid number in the response. Then you don't need to import the client

sdk/src/gateway/layerzero-client.ts
```diff
@@ -0,0 +1,104 @@
+export type LayerZeroChainInfo = {
+    name: string;
+    eid?: string;
+    oftAddress: string;
+    nativeChainId?: number;
+};
+
+export class LayerZeroClient {
```
gregdhill: Let's leave this in the `layerzero.ts` for now

sdk/src/gateway/layerzero.ts
gregdhill: Why would you have cycle imports?

gregdhill (changes requested):

sdk/src/gateway/layerzero-client.ts
mastilver: it's not possible

PR #772, opened 2025-09-19
refactor: max fees to swap for l0 strategy

Orland0x:
Adds separate fee buffers to account for the finality delays between both Bitcoin -> Bob, and Bob -> L0 Destination Chain.

The Bitcoin -> Bob buffer needs to be quite large to account for the 30 min + for BTC finality, which can cause gas price to quite significantly change between quote and execution. 

## Summary by CodeRabbit

* **New Features**
  * Added configurable origin and destination finality buffers to control fee buffering and cap the max tokens swapped for cross-chain fees.

* **API Changes**
  * Replaced single fee-buffer option with separate origin/destination finality buffer parameters.

* **Bug Fixes**
  * Quotes now use the native network fee directly and reject orders where fee swaps would consume the full amount, with a clear error.

* **Tests / Chores**
  * Tests updated to use the new buffer settings; package version bumped.

### DIFF
--- a/sdk/test/layerzero.test.ts
+++ b/sdk/test/layerzero.test.ts
@@ -75,11 +75,12 @@ describe('LayerZero Tests', () => {
         const quote = await client.getQuote({
             fromChain: 'bitcoin',
             fromToken: 'bitcoin',
-            toChain: 'base',
+            toChain: 'ethereum',
             toToken: '0x0555E30da8f98308EdB960aa94C0Db47230d2B9c',
             fromUserAddress: 'bc1q6tgkjx4pgc5qda52fsgeuvjrhml5nuawwplejq',
             toUserAddress: '0x2A7f5295ac6e24b6D2ca78d82E3cbf01dDA52745',
             amount: 3000,
+            l0FeeBuffer: 30000,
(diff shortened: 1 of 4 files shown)


sdk/test/layerzero.test.ts
```diff
@@ -11,7 +11,7 @@ import { Hex } from 'viem';
 import { privateKeyToAccount } from 'viem/accounts';
 
 describe('LayerZero Tests', () => {
-    it('should get chains', async () => {
+    it.skip('should get chains', async () => {
```
gregdhill, line 14: Revert

sdk/test/layerzero.test.ts
```diff
@@ -69,17 +69,18 @@ describe('LayerZero Tests', () => {
         assert.equal(findChain('sei')?.oftAddress, '0x0555e30da8f98308edb960aa94c0db47230d2b9c');
     }, 120000);
 
-    it.skip('should get an onramp quote and execute it', async () => {
+    it('should get an onramp quote and execute it', async () => {
```
gregdhill: Revert

PR #781, opened 2025-09-24
feat(gateway-v4): expose output tokens

mastilver:
## Summary by CodeRabbit

- New Features
  - Orders now expose output token address/amount and per-order helpers to retrieve output and consolidated token lists.

- Public API
  - Responses include token-received data and several order fields as explicit nullable values; token-retrieval helpers added/updated.

- Bug Fixes
  - Improved token/address and amount resolution when upstream fields are missing or null for consistent token data.

- Tests
  - Added coverage for multi-token orders and updated token expectations.

- Chore
  - Package version bumped.

### DIFF
--- a/sdk/package-lock.json
+++ b/sdk/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "@gobob/bob-sdk",
-    "version": "4.2.21",
+    "version": "4.3.2",
     "lockfileVersion": 3,
     "requires": true,
     "packages": {
         "": {
             "name": "@gobob/bob-sdk",
-            "version": "4.2.21",
+            "version": "4.3.2",
             "dependencies": {
                 "@bitcoinerlab/secp256k1": "^1.2.0",
                 "@eslint/eslintrc": "^3.3.1",
--- a/sdk/package.json
+++ b/sdk/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@gobob/bob-sdk",
-    "version": "4.3.1",
+    "version": "4.3.2",
     "main": "dist/index.js",
     "types": "dist/index.d.ts",
     "scripts": {


sdk/src/gateway/types.ts
```diff
@@ -286,6 +286,11 @@ export type OfframpGatewayCreateQuoteResponse = {
     gateway: Address;
 };
 
+export interface TokenReceived {
+    amount: string;
+    token_address: Address;
```
nakul1010: replace with `tokenAddress`

sdk/src/gateway/types.ts
```diff
@@ -313,15 +318,17 @@ export interface OnrampOrderResponse {
...
     /** @description The output amount (from strategies) */
-    outputTokenAmount?: string;
+    outputTokenAmount?: string | null;
     /** @description The tx hash on the EVM chain */
     txHash?: string;
     /** @description V4 order details */
     orderDetails?: OrderDetailsRaw;
     /** layerzero dst eid if the order being routed through layerzero */
     layerzeroDstEid?: number;
+    /** tokens received for gatway v4 order */
```
nakul1010: Replace with `ERC20 tokens received by the user for gateway order`

sdk/src/gateway/client.ts
```diff
@@ -1007,6 +1007,9 @@ export class GatewayApiClient {
         );
         const orders: OnrampOrderResponse[] = await response.json();
         return orders.map((order) => {
+            const outputTokenAddress = order.outputTokenAddress ?? order.tokensReceived?.[0]?.tokenAddress ?? undefined;
+            const outputTokenAmount = order.outputTokenAmount ?? order.tokensReceived?.[0]?.amount ?? undefined;
```
nakul1010: `order.tokensReceived?.[0]`
I think it should be the last token in the list and not the first one, right?

nakul1010:
Update SDK tag as well to `4.2.19`

sdk/src/gateway/client.ts
mastilver: It's just a fallback. Ideally we stop using it.

Why should it be the last one?

nakul1010: Most likely, the last token address represents what the user actually receives.
For example, if a user swaps WBTC → USDC → USDT because Uniswap doesn’t have a direct WBTC → USDT pool, the final transfer event will be for USDT.
This ensures that the last token recorded is indeed what the user ends up holding in their wallet, not just an intermediate step.

sdk/src/gateway/types.ts
```diff
@@ -313,15 +318,17 @@ export interface OnrampOrderResponse {
     /** @description The amount of ETH received */
     outputEthAmount?: string;
     /** @description The output token (from strategies) */
-    outputTokenAddress?: Address;
+    outputTokenAddress?: Address | null;
```
gregdhill: This can be undefined and null? Would keep strictly as `?: Address`

gregdhill: Same below

mastilver: I'm not sure why it was undefined in the first place. But on the server side `output_token_address` is `Option<Address>` which convert to `null` if `None`

sdk/src/gateway/client.ts
mastilver: tokensReceived only contains tokens received by the recipient. most of the time it should contains one token, it could contains more if the user receive multiple tokens

sdk/src/gateway/client.ts
```diff
@@ -1046,6 +1059,30 @@ export class GatewayApiClient {
...
+            };
+
+            const getTokens = () => {
+                const tokens = order.tokensReceived
+                    ? order.tokensReceived
+                    : [{ amount: getTokenAmount(), tokenAddress: getTokenAddress() }];
+
+                return tokens.map(({ amount, tokenAddress }) => ({
+                    amount: amount,
+                    token: ADDRESS_LOOKUP[chainId][tokenAddress],
```
nakul1010, line 1176: Since you are using `ADDRESS_LOOKUP`, we need to keep the token list up to date. Also, I hope it doesn’t fail if the API returns a new token that isn’t present in `tokens.ts`. It should still return the rest of the tokens available in `tokens.ts`.

Ideally, we should have a test case to cover this scenario.

mastilver: that's a very good points

PR #783, opened 2025-09-25
fix(onramp): simplify getStatus logic by removing outputTokenAddress check

nakul1010:


sdk/test/gateway.test.ts
```diff
@@ -1041,6 +1041,6 @@ describe('Gateway Tests', () => {
                 throw new Error('Expected gasFee to exist on feeBreakdown, but it does not.');
             }
         },
-        { timeout: 10000 } // 10 seconds
+        { timeout: 30000 } // 30 seconds
```
nakul1010, line 1044: takes time to run locally

PR #794, opened 2025-10-01
fix: gateway client getQuote types

danielsimao:
## Summary by CodeRabbit

- New Features
  - Unified quote format for onramp/offramp using a type/data wrapper across getQuote and executeQuote.
  - getQuote now returns a structured object with type, params, data, and final fee/output fields.
  - executeQuote now accepts a single wrapped quote object and validates quote type.
  - OnrampQuote adds an outputSatoshis field.

- Refactor
  - Streamlined quote handling to a discriminated type/data model for consistent flows.

- Tests
  - Updated tests to use the new wrapped quote shape.

- Chores
  - Version bumped to 4.3.0.

### DIFF
--- a/sdk/src/gateway/layerzero.ts
+++ b/sdk/src/gateway/layerzero.ts
@@ -1,12 +1,11 @@
-import { Address, encodeAbiParameters, encodePacked, Hex, padHex, parseAbiParameters } from 'viem';
-import { AllWalletClientParams, GatewayApiClient } from './client';
-import { ExecuteQuoteParams, GetQuoteParams, OfframpExecuteQuoteParams } from './types';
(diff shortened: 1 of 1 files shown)


sdk/src/gateway/client.ts
```diff
@@ -914,6 +909,8 @@ export class GatewayApiClient {
 
             return transactionHash;
         }
+
+        throw new Error('Invalid quote type');
```
danielsimao, line 913: Above a check against the expected type variable. In case the quote coming in does not have those, error is thrown.

slavastartsev (approved):

gregdhill (approved):

Orland0x (approved):

PR #800, opened 2025-10-06
chore: add blockchain fees for get quote method

slavastartsev:
## Summary by CodeRabbit

- New Features
  - Offramp quotes now include an estimated on-chain gas fee in the fee breakdown and a new public gas-cost estimation method for on-chain order creation.
  - Automatic chain-aware resolution for on-chain cost estimates.

- Behavior
  - Gas estimation is error-tolerant: failures fall back to a zero gas fee and emit warnings.

- Documentation
  - SDK types updated to expose an optional gasFee in offramp fee breakdowns.

- Tests
  - Added tests verifying gas fee presence across offramp scenarios.

### DIFF
--- a/sdk/src/gateway/client.ts
+++ b/sdk/src/gateway/client.ts
@@ -226,8 +226,13 @@ export class GatewayApiClient {
                 params,
                 finalOutputSats: data.amountReceiveInSat,
                 finalFeeSats: data.feeBreakdown.overallFeeSats,
-                data,
-                blockchainFee: createOrderGasCost,
+                data: {
+                    ...data,
+                    feeBreakdown: {
+                        ...data.feeBreakdown,
+                        gasFee: createOrderGasCost,
+                    },
+                },
             };
         }
 
--- a/sdk/src/gateway/client.ts
+++ b/sdk/src/gateway/client.ts
@@ -13,6 +13,8 @@ import {
     Chain as ViemChain,
     WalletClient,
     zeroAddress,
+    parseEther,
+    toHex,
 } from 'viem';
 import { bob, bobSepolia } from 'viem/chains';
 import { EsploraClient } from '../esplora';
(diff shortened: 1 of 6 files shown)


danielsimao:
We should rethink the current structure since both the **origin** and **destination** chains include a `blockchainFee`.

This looks like a good opportunity to **refactor** the way we represent fees.
On the UI, we display two main types of fees: **protocol** and **network**.
However, **network fees** differ depending on whether we’re performing an **onramp** or **offramp** transaction.

**Current logic:**

* **Onramp:**

  * Origin (BTC) network fee
  * Destination (EVM) network fee
  * Protocol fee

* **Offramp:**

  * Origin (EVM) network fee
  * Destination (BTC) network fee
  * Protocol fee

To make this clearer and more consistent, the final API response could follow this structure:

```ts
{
  (...),
  protocolFee: X,
  networkFee: {
    origin: Y,
    destination: Z
  }
}
```

This would make it easier to reason about both directions and unify the fee handling logic across onramp/offramp flows.

danielsimao (approved):

sdk/src/gateway/client.ts
```diff
@@ -197,13 +198,41 @@ export class GatewayApiClient {
...
+
+            const fee = feeValues.maxFeePerGas ?? gasPrice;
+
+            const createOrderGasEstimate = await publicClient.estimateContractGas({
+                address: offrampRegistryAddress,
+                abi: offrampOrder.offrampABI,
+                functionName: offrampOrder.offrampFunctionName,
+                args: offrampOrder.offrampArgs,
+                account: params.fromUserAddress as Address,
+            });
```
nakul1010: Qs: say `fromUserAddress` doesn't have enough token allowance/balance, the `estimateContractGas` will fail right? 
We should do `stateOverride`

nakul1010 (approved):

PR #813, opened 2025-10-13
refactor: override error message for L0 offramp insufficient fees

Orland0x:
## Summary by CodeRabbit

* **Improvements**
  * More robust transaction simulation handling across onramp, offramp and cross-chain swap flows for clearer outcomes.

* **Bug Fixes**
  * Low native balance failures now surface a clear "Insufficient native funds" message; other error behavior preserved.

* **New Features**
  * Expanded token list with broader asset coverage and richer metadata for improved asset support.

* **Chores**
  * Project configuration updated to include the token list asset.

### DIFF
--- a/sdk/src/gateway/layerzero.ts
+++ b/sdk/src/gateway/layerzero.ts
@@ -2,10 +2,12 @@ import ecc from '@bitcoinerlab/secp256k1';
 import * as bitcoin from 'bitcoinjs-lib';
 import {
     Address,
+    CallExecutionError,
     ContractFunctionExecutionError,
     encodeAbiParameters,
     encodePacked,
     Hex,
+    InsufficientFundsError,
     isAddress,
     isAddressEqual,
     maxUint256,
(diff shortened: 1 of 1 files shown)


sdk/src/gateway/layerzero.ts
```diff
@@ -456,19 +456,33 @@ export class LayerZeroGatewayClient extends GatewayApiClient {
...
+
+                return txHash;
+            } catch (error) {
+                const message = error instanceof Error ? error.message : String(error);
+                if (
+                    message.includes(
+                        'The total cost (gas * gas fee + value) of executing this transaction exceeds the balance of the account'
+                    )
+                ) {
+                    throw new Error(
```
danielsimao: Ideally in the future we should throw error codes to allow the UI to handle with the right copy and tranlastions

slavastartsev: ideally we need to validate which error class was used to crate an error.
you can do similar check to the way we check for user rejection event https://github.com/bob-collective/ui/blob/710a43ce4a529177117da6ce5f583509d2e6e39c/apps/evm/src/lib/viem/utils.ts#L3-L4

string matching is not the best solution as error message is a subject to change

I think for the error message you can stick to `viem` error message as those are nicely formatted, so you're safe to re-throw

Orland0x: ok slightly unsure what changes you recommend to make now @slavastartsev ? you can just make a commit yourself if easier also

slavastartsev: What I wanted to say is that viem chains the errors. Described in [this test](https://github.com/wevm/viem/blob/3aa882692d2c4af3f5e9cc152099e07cde28e551/src/actions/public/simulateContract.test.ts#L711).
the way we need to adjust the `catch` block
```typescript
} catch (error) {
    if (error instanceof ContractFunctionExecutionError) {
        throw error; // re-throw the error throw by viem or throw new Error with custom error message
    }

    throw error;
}
```

Orland0x: Is the `The total cost (gas * gas fee + value) of executing this transaction exceeds the balance of the account` error the only instance of `ContractFunctionExecutionError` that could come up?

PR #821, opened 2025-10-20
chore: remap chains

slavastartsev:
## Release Notes

* **Chores**
  * Version bumped to 4.3.9
  * Updated viem dependency to ^2.38.3
  * Added package manager specification

### DIFF
--- a/sdk/package.json
+++ b/sdk/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@gobob/bob-sdk",
-    "version": "4.3.10",
+    "version": "4.3.9",
     "main": "dist/index.js",
     "types": "dist/index.d.ts",
     "scripts": {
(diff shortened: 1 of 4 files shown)


slavastartsev:
> Why do we need this change? Goal of the L0 client is to fetch the addresses dynamically

We need this change due to the `getTokenSlots` [link](https://github.com/bob-collective/bob/blob/c63a411d99aad294dd8d94047fcb1f7134e4307a/sdk/src/gateway/tokens.ts#L586) function .
To estimate blockchain fees when the user address is not provided, we must use a dummy address and override the contract's balance and allowance values for it. Since getTokenSlots supports all tokens (not just LayerZero tokens), we need to specify WBTC OFT token configurations for every supported chain.
This is a work in progress. A better alternative would be to extend the LayerZero client by adding a getTokenSlots method that returns fixed values based on the specified token.

slavastartsev:
changes is this PR are no longer relevant. closing

PR #839, opened 2025-11-05
fix: add onramp and offramp liquidity fetch methods with unit tests

nakul1010:
## 📝 PR Description
- Add onramp and offramp liquidity methods that give min and max amount

## Requested UI Changes
- Use the new liquidity methods
- Currently, entering a large amount triggers an **“Insufficient Liquidity”** or **rate limit** error, while very small amounts may result in a **“Fee Insufficient”** error.
Screenshot 2025-11-05 at 5 26 02 PM

- Instead, I’d prefer that users never see these errors. The UI should restrict input to values between `min_amount` and `max_amount` using the new SDK liquidity methods.

Screenshot 2025-11-05 at 5 26 13 PM

- If the entered amount falls outside the given liquidity range, the corresponding cell should highlight in red. 
    -  Eg: user entered amount 1 BTC, but max allowed is 0.5. err msg in UI: `Amount to sell at max allowed {0.5}`    
    -  Eg: user entered amount 0.00001 BTC, but min allowed is 0.001. err msg in UI: `Amount to sell at min allowed {0.001}`
- Pass zero address to method if user has not connected the wallet. 

## Summary by CodeRabbit

* **New Features**
  * Added V2 offramp liquidity query and a new onramp liquidity query (onramp supports optional gasRefill); responses include sats-based fields and a nested minimum-offramp-quote.

* **Documentation**
  * Affiliate-related JSDoc updated to include affiliate fee parameters.

* **Tests**
  * Added tests for V2 offramp and onramp responses, including big-int normalization and symbol/address resolution.

* **Chores**
  * Package version bumped.

### DIFF
--- a/sdk/src/gateway/client.ts
+++ b/sdk/src/gateway/client.ts
@@ -499,7 +499,7 @@ export class GatewayApiClient extends BaseClient {
             userAddress: userAddress,
         });
 
-        const requestUrl = `${this.baseUrl}/v2/offramp-liquidity?${queryParams.toString()}`;
+        const requestUrl = `${this.baseUrl}/v2/offramp-liquidity?${queryParams}`;
         const response = await this.safeFetch(requestUrl, undefined, 'Failed to get offramp v2 liquidity');
 
         if (!response.ok) {
--- a/sdk/test/gateway.test.ts
+++ b/sdk/test/gateway.test.ts
@@ -1470,7 +1470,7 @@ describe('Gateway Tests', () => {
         'get offramp liquidity and get quote e2e',
         async () => {
             const gatewaySDK = new GatewaySDK(bob.id);
-            const tokenAddress = SYMBOL_LOOKUP[bob.id]['wbtc (oft)'].address as Address;
+            const tokenAddress = SYMBOL_LOOKUP[bob.id]['wbtc'].address as Address;
             const offrampLiquidity = await gatewaySDK.fetchOfframpLiquidity(tokenAddress, zeroAddress);
 
             const minGatewayQuoteOnly = await gatewaySDK.fetchOfframpQuote(
@@ -1505,7 +1505,7 @@ describe('Gateway Tests', () => {
                 fromChain: 'Bitcoin',
                 fromToken: 'bitcoin',
                 toChain: bob.id,
-                toToken: 'WBTC (OFT)',
+                toToken: 'wbtc',
                 toUserAddress: tokenAddress,
                 amount: onrampLiquidity.minSatsAmount,
             });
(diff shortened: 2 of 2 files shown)


sdk/src/gateway/client.ts
```diff
@@ -478,6 +480,84 @@ export class GatewayApiClient extends BaseClient {
...
 
+    /**
+     * Fetches available offramp liquidity.
+     *
+     * @param token Token symbol or address
+     * @param userAddress User address to get liquidity for
+     * @returns Promise resolving to liquidity information
+     * @throws {Error} If API request fails
+     */
+    async fetchOfframpLiquidityV2(token: string, userAddress: Address): Promise<OfframpLiquidityV2> {
```
gregdhill: Deprecate the old method and remove the V2 here?

sdk/test/gateway.test.ts
```diff
@@ -1417,4 +1417,124 @@ describe('Gateway Tests', () => {
 
         expect(orders).toEqual([]);
     });
+
+    it('should return mocked offramp liquidity v2', async () => {
```
gregdhill: I don't think these unit tests really give us a lot, but if they work we can keep them here

sdk/test/gateway.test.ts
```diff
@@ -769,30 +769,42 @@ describe('Gateway Tests', () => {
         expect(result).toBe(true);
     });
 
+    function normalizeBigInts<T>(value: T): T {
+        return JSON.parse(JSON.stringify(value, (_, v) => (typeof v === 'bigint' ? Number(v) : v)));
+    }
```
slavastartsev, line 774: feel free to skip this comment
```typescript
  Object.defineProperty(BigInt.prototype, "toJSON", {
    value: function () {
      return this.toString()
    },
    writable: true,
    configurable: true
  })
```

sdk/src/gateway/client.ts
```diff
@@ -483,33 +484,80 @@ export class GatewayApiClient extends BaseClient {
...
-            `${this.baseUrl}/offramp-liquidity/${tokenAddress}`,
-            undefined,
-            'Failed to get offramp liquidity'
-        );
+        const queryParams = new URLSearchParams({
+            tokenAddress: tokenAddress,
+            userAddress: userAddress,
+        });
+
+        const requestUrl = `${this.baseUrl}/v2/offramp-liquidity?${queryParams.toString()}`;
```
slavastartsev: NIT comment
JS implicitly calls `.toString()` for all templated values
```suggestion
        const requestUrl = `${this.baseUrl}/v2/offramp-liquidity?${queryParams}`;
```

slavastartsev (approved):

sdk/test/gateway.test.ts
nakul1010: it adds `""` to field hence `expect`  doesn't work. So keeping it as is for now

PR #848, opened 2025-11-11
chore: add tokenlist

slavastartsev:
## Release Notes

* **New Features**
  * Introduced comprehensive token registry with 40+ tokens including stablecoins, wrapped assets, and bridged variants across multiple networks.
  * Added support for cross-chain token bridges with metadata for Ethereum, BOB, and testnet deployments.
  * Implemented standardized token list schema following industry standards.

* **Documentation**
  * Updated token registry reference documentation with new token list integration patterns and bridge information structure.

### DIFF
--- a/sdk/src/assets/token-list.json
+++ b/sdk/src/assets/token-list.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://raw.githubusercontent.com/Uniswap/token-lists/refs/heads/main/src/tokenlist.schema.json",
   "name": "BOB Tokens",
-  "timestamp": "2025-11-17T20:28:31.391Z",
+  "timestamp": "2025-11-20T14:52:09.564Z",
   "version": {
     "major": 1,
     "minor": 0,
@@ -65,6 +65,42 @@
       "logoURI": "https://raw.githubusercontent.com/bob-collective/bob/refs/heads/master/tokenlist/data/BABY/logo.svg",
       "extensions": {
         "tokenId": "BABY",
+        "bridge": {}
+      }
+    },
+    {
+      "chainId": 1,
+      "address": "0xC9746F73cC33a36c2cD55b8aEFD732586946Cedd",
+      "name": "BOB",
+      "symbol": "BOB",
+      "decimals": 18,
+      "logoURI": "https://raw.githubusercontent.com/bob-collective/bob/refs/heads/master/tokenlist/data/BOB/logo.svg",
+      "extensions": {
+        "tokenId": "BOB",
+        "bridge": {}
+      }
+    },
+    {
+      "chainId": 60808,
+      "address": "0xB0BD54846a92b214C04A63B26AD7Dc5e19A60808",
+      "name": "BOB",
+      "symbol": "BOB",
+      "decimals": 18,
+      "logoURI": "https://raw.githubusercontent.com/bob-collective/bob/refs/heads/master/tokenlist/data/BOB/logo.svg",
+      "extensions": {
+        "tokenId": "BOB",
+        "bridge": {}
+      }
+    },
+    {
+      "chainId": 56,
+      "address": "0x52B5fB4B0F6572B8C44d0251Cc224513ac5eB7E7",
+      "name": "BOB",
+      "symbol": "BOB",
+      "decimals": 18,
+      "logoURI": "https://raw.githubusercontent.com/bob-collective/bob/refs/heads/master/tokenlist/data/BOB/logo.svg",
+      "extensions": {
+        "tokenId": "BOB",
         "bridge": {}
       }
     },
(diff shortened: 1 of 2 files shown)


sdk/src/scripts/verify-tokenlist.ts
```diff
@@ -8,7 +8,7 @@ import Ajv from 'ajv';
 import addFormats from 'ajv-formats';
 
 const SCHEMA_URL = 'https://raw.githubusercontent.com/Uniswap/token-lists/refs/heads/main/src/tokenlist.schema.json';
-const tokenlistPath = path.join(__dirname, '../../../token-list.json');
+const tokenlistPath = path.join(__dirname, '../../../tokenlist/token-list.json');
```
gregdhill: Move this file to tokenlist package

sdk/package.json
```diff
@@ -43,6 +43,7 @@
         "yargs": "^18.0.0"
     },
     "dependencies": {
+        "@gobob/tokenlist": "file:../tokenlist",
```
gregdhill: Will this work after we publish?

slavastartsev: ideally it should work as tokenlist just another npm dependency
I will verify if works as expected

slavastartsev: it turns out npm is not that capable as compared to `pnpm` -- it cannot resolve remote links
which means that we will need to create `build & publish` pipeline for tokenlist repo. 
This could be easily done with `changeset` bot. @gregdhill should I add it?

sdk/src/gateway/tokens.ts
```diff
@@ -102,24 +112,8 @@ const optimismTokens = [
...
-        address: '0x2260fac5e5542a773aa44fbcfedf7c193bc2c599',
-        symbol: 'WBTC',
-        decimals: 8,
-        chainId: 1,
-        logoURI: 'https://raw.githubusercontent.com/bob-collective/bob/master/assets/wbtc.svg',
-        allowanceSlot: 2n,
-        balanceSlot: 0n,
-    },
+    ethereumTokensMapping['Solv'],
+    { ...ethereumTokensMapping['WBTC'], allowanceSlot: 2n, balanceSlot: 0n },
```
gregdhill: Reuse STORAGE_SLOTS_MAP

gregdhill: Would also be good to index by address not name to be specific (for ex if there are two wBTC tokens)

.changeset/config.json
```diff
@@ -0,0 +1,11 @@
...
+  "$schema": "https://unpkg.com/@changesets/config@2.3.0/schema.json",
+  "changelog": "@changesets/cli/changelog",
+  "commit": false,
+  "fixed": [],
+  "linked": [],
+  "access": "public",
+  "baseBranch": "master",
+  "updateInternalDependencies": "patch",
+  "ignore": []
+}
```
gregdhill: Not a huge fan of these because they are almost never maintained (unless project is big enough to have many downstream consumers)

gregdhill: Also not sure about shoehorning this into an unrelated PR

.github/workflows/tokenlist-changesets.yml
```diff
@@ -0,0 +1,74 @@
+---
```
gregdhill: We should have a workflow for PRs to ensure addresses are checksummed, or add a way to ensure that happens in the generate script

gregdhill: Also to use the verify script I added to ensure all the contracts on BOB have the correct name, decimals etc.

tokenlist/types.ts
```diff
@@ -0,0 +1,44 @@
...
+import { supportedChainMapping, supportedChains } from './config';
+
+export type KebabCase<T extends string> = T extends `${infer S} ${infer E}`
+  ? `${Lowercase<S>}-${KebabCase<E>}`
+  : Lowercase<T>;
+
+export type ValueOf<T> = T[keyof T];
+export type Entries<T> = [keyof T, ValueOf<T>][];
+
+export type SuppertedChain = keyof typeof supportedChainMapping;
```
gregdhill:
```suggestion
export type SupportedChain = keyof typeof supportedChainMapping;
```

.github/workflows/tokenlist-changesets.yml
slavastartsev: > We should have a workflow for PRs to ensure addresses are checksummed, or add a way to ensure that happens in the generate script

already implemented in [generate script](https://github.com/bob-collective/bob/blob/7d62cef9b243396dbf489ed0b3672150abd58500/tokenlist/scripts/generate.ts#L46)

sdk/package.json
gregdhill: Would just switch to pnpm if you want

slavastartsev: pnpm is not gonna help
any project that uses sdk as a dependency is not gonna able to resolve this dependency as it lives outside of the sdk directory

PR #851, opened 2025-11-13
feat: chain name resolving

Orland0x:
## Release Notes

* **Improvements**
  * Streamlined LayerZero gateway initialization—no longer requires explicit chain parameters at startup
  * Enhanced chain resolution to better support multiple blockchain networks
  * Improved error messaging to clearly display supported chain names
  * Optimized cross-chain quote and execution workflows

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #875, opened 2025-11-25
fix: create `getTokenDetails` method instead of direct look up

nakul1010:
## Summary by CodeRabbit

* **Chores**
  * Bumped SDK version to 4.4.3
  * Updated formatting ignore list to exclude asset files

* **Bug Fixes**
  * Improved token resolution with more robust fallback and graceful handling when tokens are unknown
  * Aligned tests and mocks to use token address values for more accurate test coverage

### DIFF
--- a/sdk/src/gateway/client.ts
+++ b/sdk/src/gateway/client.ts
@@ -1136,11 +1136,7 @@ export class GatewayApiClient extends BaseClient {
                 return getFinal(amount, outputTokenAmount);
             };
             const getToken = (): Token | undefined => {
-                try {
-                    return getTokenDetails(chainId, getTokenAddress());
-                } catch {
-                    return undefined;
-                }
+                return getTokenDetails(chainId, getTokenAddress());
             };
             const getConfirmations = async (esploraClient: EsploraClient, latestHeight?: number) => {
                 const txStatus = await esploraClient.getTransactionStatus(order.txid);


sdk/src/gateway/tokens.ts
```diff
@@ -204,3 +204,8 @@ export function getTokenSlots(tokenAddress: Address): { allowanceSlot: bigint; b
 
     return slots;
 }
+
+export function getTokenDetails(chainId: number, token: string): Token {
+    const tokenAddress = getTokenAddress(chainId, token);
+    return ADDRESS_LOOKUP[chainId][tokenAddress];
+}
```
slavastartsev: I don't see if `token` argument provided as token name/symbol so it's safe to apply constrain that token is Address. in this case it all could be simplified to
```suggestion
export function getTokenDetails(chainId: number, token: Address): Token {
    return ADDRESS_LOOKUP[chainId][getAddress(tokenAddress)];
}
```

sdk/.prettierignore
```diff
@@ -1,2 +1,3 @@
 dist
 node_modules
+src/assets/**
```
slavastartsev:
```suggestion
src/assets/
```
this will conflict with https://github.com/bob-collective/bob/pull/851/files

sdk/src/gateway/strategy.ts
```diff
@@ -250,20 +250,12 @@ export default class StrategyClient {
...
         return tokens
-            .map(
-                (addr) =>
-                    ADDRESS_LOOKUP[bob.id]?.[addr.toLowerCase()] ??
-                    ADDRESS_LOOKUP[optimism.id]?.[addr.toLowerCase()] ??
-                    ADDRESS_LOOKUP[mainnet.id]?.[addr.toLowerCase()]
-            )
-            .filter((t): t is Token => Boolean(t));
+            .map((t) => getTokenDetails(bob.id, t) || getTokenDetails(optimism.id, t) || getTokenDetails(mainnet.id, t))
+            .filter((tok): tok is Token => tok !== null);
```
slavastartsev: not sure if have `null` values mapped to `address` key in `ADDRESS_LOOKUP`.
function signature is `getTokenDetails(chainId: number, token: string): Token | undefined`
both `Token` and `undefined` not equal to `null`
```suggestion
            .map((t) => getTokenDetails(bob.id, t) || getTokenDetails(optimism.id, t) || getTokenDetails(mainnet.id, t))
            .filter((tok): tok is Token => Boolean(t));
```

sdk/src/gateway/client.ts
```diff
@@ -1137,7 +1137,11 @@ export class GatewayApiClient extends BaseClient {
                 return getFinal(amount, outputTokenAmount);
             };
             const getToken = (): Token | undefined => {
-                return ADDRESS_LOOKUP[chainId][getTokenAddress()];
+                try {
+                    return getTokenDetails(chainId, getTokenAddress());
+                } catch {
+                    return undefined;
+                }
```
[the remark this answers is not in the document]
slavastartsev: AI is correct

sdk/src/gateway/strategy.ts
slavastartsev: or even like this
```typescript
.filter(Boolean) as Token[];
```

PR #877, opened 2025-11-26
fix: check allowance & approve on eth mainnet

slavastartsev:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of ETH-origin cross-chain transfers by automatically checking and managing token approvals for wrapped assets before sending, reducing approval-related failures.
  * Ensures pre-send approval steps run and complete prior to transaction submission so transfers proceed more consistently and predictably.

### DIFF
--- a/sdk/src/gateway/layerzero.ts
+++ b/sdk/src/gateway/layerzero.ts
@@ -51,7 +51,8 @@ bitcoin.initEccLib(ecc);
 /**
  * L0 recipient that forwards messages to create offramp orders on-chain.
  */
-export const OFFRAMP_COMPOSER = '0xDeEbEed29d8aeF2610fa924bc87fd1C9DF94e655';
+export const OFFRAMP_COMPOSER = '0xaffBF9ECC4a23adfFe887FB859654B8B780CCed0';
+export const LAYERZERO_STRATEGY = '0x4572ce66cB33255B60a15e3c6cb2ef9c65A30ebC';
 
 export class LayerZeroClient {
     private basePath: string;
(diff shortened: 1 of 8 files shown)


sdk/src/gateway/layerzero.ts
```diff
@@ -572,6 +572,33 @@ export class LayerZeroGatewayClient extends GatewayApiClient {
                 };
 
                 try {
+                    // ETH -> %any_chain%
+                    if (data.sourceEid === 30101) {
```
nakul1010, line 756: Qs: Why just for eth and not other chain like optimism

sdk/src/gateway/layerzero.ts
```diff
@@ -572,6 +572,33 @@ export class LayerZeroGatewayClient extends GatewayApiClient {
...
+                        const wbtcMainnetAddress = getTokenAddress(mainnet.id, 'wbtc');
```
nakul1010, line 757: I think you can skip this and just use the same oftAddress, since the LayerZero WBTC address is the same on both ETH and BOB.

nakul1010: @Orland0x to confirm

sdk/src/gateway/layerzero.ts
slavastartsev: I'm not sure if optimism has OFTAdapter contract deployed

sdk/src/gateway/layerzero.ts
slavastartsev: the trick with l0 is that oft contract on mainnet is not extending erc20 interface but is OFTAdapter. on any other chain we are no longer dealing with OFTAdapter but `%ERC20%OFT` which extends erc20.
[OFTAdapter](https://github.com/LayerZero-Labs/devtools/blob/de203c58a2064092d1f8fc141c777cb44b8db76b/packages/oft-evm/contracts/OFTAdapter.sol#L20)
[OFT token](https://github.com/LayerZero-Labs/devtools/blob/de203c58a2064092d1f8fc141c777cb44b8db76b/packages/oft-evm/contracts/OFT.sol#L12)

nakul1010: Leave it got it [WBTC](https://explorer.gobob.xyz/address/0x0555E30da8f98308EdB960aa94C0Db47230d2B9c) is native token

sdk/src/gateway/layerzero.ts
nakul1010: ~~I think we need to do the same thing for optimism as well: https://metadata.layerzero-api.com/v1/metadata/experiment/ofts/list?symbols=WBTC~~
The approvalRequired says only for eth

PR #932, opened 2026-02-03
fix: offramp `executeQuote`

nakul1010:


sdk/src/gateway/client.ts
```diff
@@ -229,87 +229,47 @@ export class GatewayApiClient {
...
-                    address: tokenAddress,
-                    abi: erc20Abi,
-                    functionName: 'approve',
-                    args: [spenderAddress, maxUint256],
-                });
-
-                const approveTxHash = await walletClient.writeContract(request);
-
+                let approveTxHash: Hash;
+                // Skip simulateContract for USDT
```
sander2: why skip simulation?

sdk/src/gateway/client.ts
```diff
@@ -229,87 +229,47 @@ export class GatewayApiClient {
...
+                if (isAddressEqual(tokenAddress, ETHEREUM_USDT_ADDRESS)) {
```
slavastartsev: this logic applies not only to USDT but many other old contracts. Need to make logic generic

PR #935, opened 2026-02-04
fix: return error message instead of status text

nakul1010:


sdk/src/gateway/generated-client/runtime.ts
```diff
@@ -262,12 +262,15 @@ export class ResponseError extends Error {
     readonly body: unknown;
 
     constructor(response: Response, body: unknown = undefined) {
-        const message =
-            (body as any)?.message ??
-            `${response.status} ${response.statusText}`;
+        // Prioritize the actual error message from body, fallback to status
```
slavastartsev, line 265: I doubt this would work
changes made to generated files

PR #947, opened 2026-02-19
Instant swap SDK

slavastartsev:


sdk/src/gateway/client.ts
```diff
@@ -378,6 +378,229 @@ export class GatewayApiClient {
...
+     * Executes a complete quote (onramp or offramp) in a single transaction.
+     *
+     * For onramp: creates order, signs Bitcoin transaction, and finalizes
+     * For offramp: approves tokens if needed and creates on-chain order
+     *
+     * @param params Parameters including quote and wallet clients - see {@link GatewayQuote} & {@link AllWalletClientParams}
+     * @returns Promise resolving to transaction hash
+     * @throws {Error} If required signers are missing or transaction fails
+     */
+    async _executeQuote(
```
gregdhill, line 413: Why did you add this? Looks like mostly a copy/paste so I'd like not to have this

slavastartsev: We had to add it based on conversation between me, @sander2 and @danielsimao. /get-order api only accepts (temporarily) orderId, not txId/txHash. So we had to extend SDK api with another  executeQuote method with different signature which returns both txHash and orderId. 

planning to remove this when /get-order api could handle both orderId and txHash/txId

Makefile
```diff
@@ -11,7 +11,7 @@ BINDINGS_FOLDER=bindings
 BINDINGS_CRATES_FOLDER=$(CRATES_FOLDER)/$(BINDINGS_FOLDER)
 BINDINGS_OUT_PATH=$(CONTRACTS_PATH)/out/$(BINDINGS_FOLDER)
 
-export OPEN_API_SPEC_URL="https://gateway-api-mainnet.gobob.xyz/api-doc.json"
+export OPEN_API_SPEC_URL="https://gateway-api-zeroconf.gobob.xyz/api-doc.json"
```
slavastartsev: undo

PR #1039, opened 2026-04-15
feat(sdk): add apiKey aparm to the sdk constructor

Nuhvi:


sdk/src/gateway/client.ts
```diff
@@ -106,23 +106,43 @@ export type ExecuteQuoteResult =
...
+     *
+     * // With API key
+     * const authenticatedClient = new GatewayApiClient(null, '0x1234...');
      * ```
      */
-    constructor(basePath?: string) {
+    constructor(basePath?: string, apiKey?: string) {
+        if (apiKey && apiKey.length !== 66) {
+            throw new Error('apiKey must be exactly 66 characters');
+        }
```
gregdhill, line 140:
```suggestion
    constructor(basePath?: string, apiKey?: string | Uint8Array) {
        if (apiKey) {
            const bytes = typeof apiKey === 'string'
                ? hexToBytes(apiKey)  // handles both "0x..." and raw hex
                : apiKey;

            if (bytes.length !== 32) {
                throw new Error(
                    'apiKey must be 32 bytes. Provide hex (with or without 0x prefix) or a Uint8Array.'
                );
            }
        }
```

Nuhvi: api keys are never bytes, it is an opaque string

gregdhill (approved):

PR #1042, opened 2026-04-23
chore: generate v2 api

slavastartsev:


gateway-cli/src/util/quote-v2.ts
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * V2 equivalent of the SDK's `getInnerQuote`. Picks the active variant
+ * (onramp / offramp / layerZero) from a `GatewayQuoteV2` discriminated union.
+ *
+ * The SDK only ships a typed helper for V1 quotes, but the discriminator
+ * shape is identical, so we reproduce it here against the V2 types.
+ */
+export type InnerQuoteV2 = GatewayOnrampQuoteV2 | GatewayOfframpQuoteV2 | GatewayLayerZeroQuote;
+
+export function getInnerQuoteV2(quote: GatewayQuoteV2): InnerQuoteV2 {
```
danielsimao, line 12: shouldnt we use the instance function check?

gateway-cli/src/commands/swap.ts
```diff
@@ -120,15 +121,20 @@ export async function handleSwap(opts: SwapOptions, log: Logger): Promise<SwapRe
   const { orderId, order, outputAmount } = await pRetry(async () => {
     try {
       const quote = await sdk.getQuote(quoteParams);
-      const order = await sdk.api.createOrder({ gatewayQuote: quote });
-      const orderData = (order as any)[variant];
+      const order: GatewayCreateOrder = await getApi().createOrderV2({ gatewayQuoteV2: quote });
+      // Cross-check that the API returned the variant our chain detection expected.
+      const orderData =
+        variant === "onramp" && "onramp" in order ? order.onramp
```
danielsimao, line 127: redudant variant check no?

danielsimao (approved):

gateway-cli/src/util/quote-v2.ts
slavastartsev: quote is a plain object
not a class instance

gateway-cli/src/commands/swap.ts
slavastartsev: no

PR #1050, opened 2026-05-07
chore: use getquote and execute quote

slavastartsev:


gateway-cli/src/commands/swap.ts
```diff
@@ -1,15 +1,20 @@
...
 import { fetchPrice } from "../util/price-oracle.js";
 import { loadConfig, getSdk, getApi } from "../config.js";
-import { deriveAddress, resolveSigner, buildRegisterPayload, getChainFamily, resolvePrivateKey, resolveRecipient } from "../chains/index.js";
+import { deriveAddress, resolveSigner, getChainFamily, resolvePrivateKey, resolveRecipient } from "../chains/index.js";
 import type { Logger, SwapSuccessJson, SwapSubmittedJson, SwapMempoolPendingJson } from "../output.js";
 
+// Default sender for BTC onramp --unsigned without --sender or BITCOIN_PRIVATE_KEY.
+// The gateway needs a sender to construct the PSBT; this address acts as a stand-in
+// when the caller will substitute their own inputs externally.
+const WALLETLESS_BTC_SENDER = "bc1qxs6ndmct2xhrhpzcmurnfuemhm83asxf4e2dsf";
```
nakul1010: I don't think we should have this behavior.
Let's fail instead and ask the user to provide the address.
I'm okay not having walletless flow in gateway-cli

slavastartsev: Ok, will fix that
Thanks 🙏

nakul1010 (approved):

nakul1010 (approved):

PR #1053, opened 2026-05-12
chore: sdk version 5.5.1

slavastartsev:


nakul1010 (approved):

nakul1010 (commented):
nit: maybe add it as a constant instead of harcoding at each place

PR #1060, opened 2026-05-20
feat: price impact + Token swap

slavastartsev:


nakul1010 (approved):
Could you update package.json to release a stable SDK version and use that in UI? 
Otherwise LGTM

slavastartsev:
> Could you update package.json to release a stable SDK version and use that in UI? Otherwise LGTM

version 5.5.3 is specified in package.json file

PR #1064, opened 2026-05-22
chore: add callback property to execute quote method

slavastartsev:


danielsimao (approved):

danielsimao:
# PR Review Summary

Comprehensive review across code quality, tests, types, and error handling. Scope is SDK-only.

## Critical Issues (must fix before merge)

1. **Lint fails on the new code** — `pnpm run lint` errors on `sdk/src/gateway/client.ts:38` and `:229` (prettier multi-line violations). Run `pnpm run format`.

2. **Unprotected callback can abort a financial transaction mid-flow** (`client.ts:251, 263, 343, 356, 361, 436, 449, 467`). A throwing consumer callback (React unmount, analytics SDK, typo) propagates and kills `executeQuote`. Worst case at lines 361/467: callback runs *after* a successful approval `waitForTransactionReceipt` but *before* `sendTransaction` — the user has paid gas to approve `maxUint256` but the swap/offramp tx never fires. Wrap each invocation in `try/catch` and log; a progress hook must not break the operation it's reporting on.

3. **Onramp `sendBitcoin` callback path has zero test coverage** (`client.ts:263`). Only `signAllInputs` is tested. Major wallets (OKX) use `sendBitcoin` — this is the dominant production path.

4. **TokenSwap (LayerZero) USDT 3-step reset+approval path is untested** (`client.ts:448`). The offramp 3-step test does NOT transitively cover this — the tokenSwap branch has its own `try/catch` (line 436) that could swallow the `ResetApproval` callback.

## Important Issues (should fix)

5. **Onramp under-reports steps relative to offramp/tokenSwap** (`client.ts:251, 263`). Both onramp paths emit `totalSteps: 1`, but `this.api.registerTxV2(...)` runs after as a meaningful async step that consumers want to display. Inconsistent with the other branches.

6. **No "callback fires before action" semantics documented**. Consumers will misinterpret a "step 2 of 3 — Approving" signal as completion. If `writeContract` throws, the UI is stuck on "Approving" forever. Either emit pre/post events with a `status` field, or document explicitly in JSDoc.

7. **TypeScript `enum` is inconsistent with the rest of the hand-written SDK** (`sdk/src/gateway/types/execute.ts`). The SDK consistently uses `as const` objects (`supportedChainsMapping`) and string-literal unions (`BitcoinNetworkName`). `enum` emits runtime code, has CJS/ESM interop quirks, and forces consumers to do a runtime import instead of `import type`. Switch to:
   ```ts
   export const ExecuteQuoteStepType = { ... } as const;
   export type ExecuteQuoteStepType = typeof ExecuteQuoteStepType[keyof typeof ExecuteQuoteStepType];
   ```
   Essentially free *before* the SDK is published; becomes a breaking change after.

8. **Type invariants are unexpressed.** `step: number` allows `0`, negatives, floats, `NaN`. `{ step: 5, totalSteps: 2 }` typechecks. A discriminated union over `type` (where `SignBitcoinTransaction` carries only `{ totalSteps: 1 }`, etc.) would make illegal states unrepresentable and would force the `step: needsReset ? 2 : 1` math at lines 356/449 to be correct at compile time.

9. **Test order is not asserted.** No callback test verifies the callback fires *before* the corresponding wallet call. If a future refactor swaps `callback?.(...)` with `await writeContract(...)`, every test still passes but the UX breaks. Use `vi.fn()` with a shared event log.

10. **Stale `needsReset`/`needsApproval` in tokenSwap** (`client.ts:412 vs 414`). `totalSteps` is captured *before* `createOrderV2`; if allowance state changes between the read at line 399 and the actual transaction, the SDK reports wrong `totalSteps` and may even skip a required reset, causing a revert that gets remapped to the misleading "Insufficient native funds" message at line 457.

## Suggestions

- **Naming:** `ExecuteQuoteStep` describes a transition event, not a step entity — `ExecuteQuoteProgress` or `ExecuteQuoteEvent` reads better. Consider prefixing with `Gateway` (matches `GatewayQuote*`, `GatewayError*`).
- **DRY:** `step: needsReset ? 2 : 1` is duplicated at lines 356 and 449. Extract a helper to keep the two branches from drifting.
- **Add a no-reset-non-USDT test** to pin the multi-condition `needsReset` predicate (`client.ts:339, 422`).
- **Pin callback-throws behavior with a test** so the chosen contract (abort vs swallow) doesn't silently regress.
- **Move `callback?.(...)` at line 449 outside the surrounding `try/catch`** so a buggy callback doesn't get remapped to a `ContractFunctionExecutionError`.

## Strengths

- All 3 `totalSteps` values (1/2/3) exercised for offramp.
- Each test asserts the full `{step, type, totalSteps}` object via `toEqual`, not just call count.
- Walletless onramp negative path explicitly tested.
- New types properly re-exported via `sdk/src/gateway/types/index.ts`.
- Snake_case enum values are stable identifiers, suitable for analytics/i18n.
- Optional callback with `?.()` keeps the API back-compatible.

## Recommended Action

1. **Block merge until:** lint passes, callback wrapped in try/catch (or contract explicitly documented), `sendBitcoin` onramp + tokenSwap USDT 3-step tests added.
2. **Strongly recommended before tagging `5.5.4`:** swap `enum` → `as const` (breaking change later), align onramp step counting with the other branches.
3. **Then:** add order-assertion test pattern, polish docstrings, address the type-design suggestions.

PR #1077, opened 2026-06-11
fix(docs): serve supported-routes in geoblocked regions + add missing chains

nud3l:
## Summary

Two fixes to the Gateway **Supported Routes & Assets** docs page (`docs.gobob.xyz/gateway/supported-routes`):

### 1. Make the page work inside geoblocked regions
The page fetched the routes list **client-side** from `gateway-api-mainnet.gobob.xyz`, which geoblocks US/OFAC regions by the *caller's* IP — so visitors in blocked regions saw "Routes could not be loaded."

- Add a Vercel serverless proxy `docs/api/get-routes.ts`, pinned to **Frankfurt (`fra1`)** via `regions` in `vercel.json`, so the gateway API sees the function's IP (allowed) instead of the visitor's. Same origin → no CORS.
- The MDX now fetches the same-origin `/api/get-routes`; tokenlist fetch stays client-side (not geoblocked).
- Proxy rejects non-GET, logs upstream failures, returns `502` on error. Guard added so a non-array response cleanly shows the error state.

### 2. Resolve symbols + explorer links for the four missing chains
`arbitrum`, `hyperliquid`, `plasma`, and `polygon` appear in the live routes feed but were absent from the page's `CHAIN_IDS` / `EXPLORERS` maps, so their tokens rendered as bare truncated addresses. Added the four entries to each map (verified every route token resolves a symbol in the tokenlist under its standard chain ID).

## Notes
- The maps remain hand-maintained: the routes API returns chain *slugs* not chain IDs, and neither viem nor the tokenlist's published JSON supplies the slug→chainId + explorer data. Eliminating the maps would require the gateway API to return `chainId` + `explorerUrl` per chain (not done here).

## Test Plan
- [ ] Preview deploy: `GET /api/get-routes` returns the routes array (from `fra1`).
- [ ] Page renders all route tables in a non-blocked region.
- [ ] Tables now load in a blocked region (VPN) where they previously failed.
- [ ] Arbitrum/Hyperliquid/Plasma/Polygon rows show symbols (USDC/WBTC/USD₮0/USDT0) and working explorer links.
- [ ] Confirm the `docs` Vercel project plan supports region pinning (Pro/Enterprise).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,3 +55,6 @@ Cargo.lock
 # Superpowers / plans
 docs/plans/
 docs/superpowers/
+
+# Vercel
+.vercel
--- a/docs/api/get-routes.ts
+++ b/docs/api/get-routes.ts
@@ -1,6 +1,10 @@
 const UPSTREAM = "https://gateway-api-mainnet.gobob.xyz/v2/get-routes";
 
 export default async function handler(req: any, res: any) {
+  if (req.method !== "GET") {
+    res.status(405).json({ error: "Method not allowed" });
+    return;
+  }
   try {
     const upstream = await fetch(UPSTREAM);
     if (!upstream.ok) {
@@ -9,7 +13,8 @@ export default async function handler(req: any, res: any) {
     }
     const data = await upstream.json();
     res.status(200).json(data);
-  } catch {
+  } catch (err) {
+    console.error("get-routes proxy error:", err);
     res.status(502).json({ error: "Failed to fetch routes" });
   }
 }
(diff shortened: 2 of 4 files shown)


nud3l:
Superseded by #1078. The geoblock will be fixed by adding an API-side exception (un-geoblocking the public routes endpoint) rather than a Vercel proxy, so the proxy work here is dropped. The symbols + explorer-links fix (the only part still needed) is cherry-picked into #1078 as a clean, focused change.

PR #1075, opened 2026-06-09
feat: add v3 api

slavastartsev:

PR #1086, opened 2026-06-22
fix(sdk): pass account object (not address) to EVM signing calls

nud3l:
## Problem

`executeQuote` (offramp / tokenSwap) and `executeStrategy` dispatch their EVM transactions with:

```js
walletClient.sendTransaction({ account: walletClient.account.address, ... })
```

In viem, a bare **address string** is a `json-rpc` account → `eth_sendTransaction` (the node/wallet signs). The **account object** from `privateKeyToAccount` is `type: 'local'` → viem signs locally → `eth_sendRawTransaction`.

Passing `.address` coerces **every** caller to json-rpc signing. That's invisible to browser wallets (they're json-rpc anyway, so the wallet signs), but it breaks **local-key callers (gateway-cli / bots)**: the local account is downgraded, viem calls `eth_sendTransaction`, and RPCs that don't support it (Tenderly) reject with `not supported`. This took down **every offramp/tokenSwap leg** of the gateway-bot once the earlier Tron crash ([#1084](https://github.com/bob-collective/bob/pull/1084)) stopped masking it — onramps (BTC→EVM, which sign a PSBT) were unaffected, matching the observed failure pattern.

## Why now

[#1075](https://github.com/bob-collective/bob/pull/1075) (v3 API) changed all these sites from `account: walletClient.account` → `account: walletClient.account.address`. The UI never noticed; the gateway-cli is the new local-account consumer.

## Fix

A bare revert to the object would be too broad: the UI's custom Tron `walletClient` ([`use-tron-wallet-client.ts`](https://github.com/bob-collective/ui/blob/44bb02aaf972f6aa858695d01615b9dd1ba34664/apps/evm/src/app/%5Blang%5D/swap/hooks/use-tron-wallet-client.ts#L153)) stores its address as **base58** with `type: 'json-rpc'`, and deliberately receives the address string (its `sendTransaction`/`writeContract` ignore the `account` field; base58↔hex is handled around `triggersmartcontract` with `visible: true`).

So the fix gates on **`account.type`** via a `signerAccount()` helper:

| Account type | Caller | Passed to viem | Result |
|---|---|---|---|
| `local` | gateway-cli / bots (private key) | account **object** | local sign → `eth_sendRawTransaction` ✅ fixed |
| `json-rpc` | browser wallets **and** the Tron client | address **string** | unchanged — wallet signs, base58↔hex handled by the client |

Applied at all 8 call sites (offramp + tokenSwap + executeStrategy). The Tron/browser path is byte-for-byte unchanged.

## Verification

- **New test**: a `type: 'local'` account object reaches `sendTransaction` on offramp (failed before the fix, passes after).
- **Companion test**: a `type: 'json-rpc'` account still receives the address string — locks in that browser/Tron behaviour is unaffected.
- Full SDK suite: **69 passing** (was 67), prettier + `tsc` clean.
- Version bumped 5.7.1 → 5.7.2.

> Not published / not rippled to gateway-cli + bot yet — opening for review first.

### DIFF
--- a/sdk/package.json
+++ b/sdk/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@gobob/bob-sdk",
-    "version": "5.7.1",
+    "version": "5.7.2",
     "main": "dist/index.js",
     "types": "dist/index.d.ts",
     "repository": {
--- a/sdk/src/gateway/client.ts
+++ b/sdk/src/gateway/client.ts
@@ -46,6 +46,23 @@ import { formatBtc, isValidTronAddress, tronAddressToHex } from './utils';
 
 const RETRY_COUNT = 8; // Number of times to retry fetching transaction receipt after sending a transaction
 
+/**
+ * Resolve the `account` to pass to viem write/simulate/send calls.
+ *
+ * Local accounts (e.g. gateway-cli / bots holding a private key) must be passed
+ * as the account OBJECT so viem signs locally (`eth_sendRawTransaction`).
+ * Passing their `.address` downgrades them to a json-rpc account, forcing
+ * `eth_sendTransaction`, which key-signing callers and non-EIP-1193 RPCs
+ * (e.g. Tenderly) reject with "not supported".
+ *
+ * json-rpc accounts keep using the address string — unchanged behaviour for
+ * browser wallets (the connected wallet signs) and the UI's custom Tron client,
+ * which stores a base58 address and handles base58<->hex conversion itself.
+ */
+function signerAccount(walletClient: WalletClient<Transport, ViemChain, Account>): Account | Address {
+    return walletClient.account.type === 'local' ? walletClient.account : walletClient.account.address;
+}
+
 export const ETHEREUM_USDT_ADDRESS = '0xdAC17F958D2ee523a2206206994597C13D831ec7';
 
 /**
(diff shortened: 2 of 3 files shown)


slavastartsev (approved):

nud3l:
Reworked this after the Tron context — thanks for flagging, the blanket revert was too broad.

I traced the UI's Tron `walletClient` ([`use-tron-wallet-client.ts`](https://github.com/bob-collective/ui/blob/44bb02aaf972f6aa858695d01615b9dd1ba34664/apps/evm/src/app/%5Blang%5D/swap/hooks/use-tron-wallet-client.ts#L153)). Confirmed: its account is `{ address: <base58>, type: 'json-rpc' }`, and its custom `sendTransaction`/`writeContract` ignore the `account` field (they use the closure `tronAddress` as `from` and call `triggersmartcontract` with `visible: true`). So the discriminator is **`account.type`**, not the address format.

Instead of reverting all sites to the object, I gate on type via a `signerAccount()` helper:

- **`local`** (gateway-cli / bots) → account **object** → viem signs locally (`eth_sendRawTransaction`). Fixes the Tenderly `eth_sendTransaction not supported` failure.
- **`json-rpc`** (browser wallets **and** the Tron client) → **address string** → byte-for-byte unchanged. The wallet signs and handles base58↔hex exactly as today.

The Tron path is untouched. Added a companion test asserting json-rpc accounts still receive the address string, so the browser/Tron behaviour is locked in. Full SDK suite green (69 passing), prettier + tsc clean.

PR #1090, opened 2026-06-23
fix(sdk): buffer gas limit on offramp/tokenSwap sends for local-key signers (#1088)

### LINKED ISSUE #1088: bob-sdk: offramp/tokenSwap send has no gas-limit buffer → out-of-gas on gas-heavy routes (velora)
## Summary
`GatewayApiClient.executeQuote` sends the offramp / tokenSwap EVM transaction **without an explicit `gas` limit**, so viem uses a bare `eth_estimateGas` with **no safety buffer**. On gas-heavy / variable routes (notably **velora** aggregator swaps), the real execution exceeds the estimate and the tx reverts **out of gas** — the order then fails silently as `{"failed":{}}`.

## Evidence (mainnet)
gateway-bot order `29cb9b44-3e46-4983-982b-e6dde9a493d2` (`USDC@ethereum → BTC`, velora) → on-chain tx [`0x5ab74cf6d09e807b3b065692d6764542e5653ceb8f93e8fb5bbed9d3e270772f`](https://etherscan.io/tx/0x5ab74cf6d09e807b3b065692d6764542e5653ceb8f93e8fb5bbed9d3e270772f):

| Field | Value |
|---|---|
| from | `0xAF91558Ba2B1994530c9cfCcbda5AE9cD2b456bb` (the bot) |
| to | `0x2e77b55…` (gateway) |
| gas limit | 1,074,362 |
| gas used | 1,071,692 (**99.8% of limit**) |
| status | `0x0` (failed) |

Tenderly trace shows OOG in a nested call (`0.1.6.1.24`) — the classic "sub-call ran out of gas because the limit was too tight" signature. The wallet had ample ETH; the **gas *limit*** was the problem.

## Root cause
`sdk/src/gateway/client.ts` — the offramp and tokenSwap sends pass `data` / `to` / `value` but no `gas`:

```js
const transactionHash = await walletClient.sendTransaction({
  account: signerAccount(walletClient),
  data: order.offramp.tx.data,
  to: spenderAddress,
  value: BigInt(order.offramp.tx.value || 0),
  // no gas → viem's raw eth_estimateGas, no headroom
});
```

`eth_estimateGas` can under-estimate aggregator swaps whose on-chain path/gas varies with market state, leaving no room for the nested swap call.

## Proposed fix
Set an explicit `gas` with headroom on the offramp / tokenSwap sends (and the approval writes): estimate gas, then apply a buffer (e.g. `estimate * 13n / 10n`, or `estimate + fixed cushion`). If `order.*.tx` already carries a backend-provided gas limit, prefer it (with buffer) over the raw estimate.

## Impact
Affects any private-key / programmatic caller (gateway-bot, server flows). The UI is less exposed (the injected wallet often pads gas). Without the buffer, gas-heavy routes intermittently OOG and surface as silent `{"failed":{}}` order failures.

Related: gateway-bot#596 (the silent-failure alert), and the recent `signerAccount` fix touched the same send sites.

nud3l:
Closes #1088.

## Problem
`GatewayApiClient.executeQuote` sends the offramp / tokenSwap EVM transactions **without an explicit `gas` limit**, so viem uses a bare `eth_estimateGas` with no safety buffer. On gas-heavy / variable routes (notably **velora** aggregator swaps) the real execution exceeds the estimate and the tx reverts out of gas — surfacing as a silent `{"failed":{}}` order.

The backend `tx` object carries no gas field, so the SDK must estimate and buffer itself. The observed failure (mainnet tx `0x5ab74cf6…0772f`) used 99.8% of a 1,074,362 limit and OOG'd in a **nested call** — the classic 63/64-rule signature where a buffered top-level limit also enlarges the gas forwarded to the sub-call.

## Fix
Add a gas buffer to the two swap `sendTransaction` sites, **for local-key signers only** (`account.type === 'local'` — gateway-bot, gateway-cli, server flows):

```
gas = max(estimate * 12n / 10n, estimate + 300_000n)
```

This `max(1.2x, +300k)` combines the two conventions already in bob-gateway: velora's `estimated_gas * 120/100` (`crates/velora/src/api/client.rs`) and the intents fixed `+300_000` cushion (`crates/intents/src/constant.rs`). Small txs get the fixed floor; large velora txs get the multiplier.

If `eth_estimateGas` reverts, the helper returns `undefined` and the send proceeds exactly as today — **no new failure mode**.

## Impact
- **gateway-bot / gateway-cli** (local key): get the buffer → OOG fixed. Picks up via an `@gobob/bob-sdk` version bump.
- **UI** (injected/json-rpc wallet): branch is skipped → behavior is **byte-for-byte unchanged** (the wallet does its own padding, as before).
- **ERC20 approvals**: out of scope — fixed/predictable gas, never the OOG culprit.

A gas *limit* is not a gas *cost* (unused gas is refunded), so erring generous is nearly free for a bot while erring tight causes the silent failures.

## Tests
`sdk/test/gateway.test.ts` — 4 new tests (offramp happy-path buffer math, estimate-throws fallback, json-rpc path unchanged, tokenSwap happy-path). Full suite: 73 passed / 11 skipped.

## Reviewed
Self-run ducklings methodology (`/bob-review`): no P0–P2 findings; the one coverage gap (tokenSwap test) was added before opening this PR.

### DIFF
--- a/sdk/src/gateway/client.ts
+++ b/sdk/src/gateway/client.ts
@@ -66,12 +66,10 @@ function signerAccount(walletClient: WalletClient<Transport, ViemChain, Account>
 /**
  * Gas-limit buffer for offramp / tokenSwap sends (bob#1088).
  *
- * Combines the two buffers bob-gateway already uses: velora multiplies its
- * estimate by 1.2 (crates/velora/src/api/client.rs) and the intents path adds a
- * fixed 300k cushion (crates/intents/src/constant.rs). We take the max so small
- * txs get the fixed floor and large (velora) txs get the multiplier. A gas
- * *limit* is not a gas *cost* — unused gas is refunded — so erring generous is
- * nearly free, while erring tight causes silent out-of-gas order failures.
+ * Takes the max of a 1.2x multiplier and a fixed 300k cushion, so small txs get
+ * the fixed floor and large (aggregator) txs get the multiplier. Unused gas is
+ * refunded, so a generous limit is nearly free while a tight one causes
+ * out-of-gas failures on gas-heavy routes.
  */
 const GAS_BUFFER_NUM = 12n;
 const GAS_BUFFER_DEN = 10n;
--- a/sdk/src/gateway/client.ts
+++ b/sdk/src/gateway/client.ts
@@ -42,7 +42,7 @@ import {
     type GetQuoteParams,
     type StrategyParams,
 } from './types';
-import { formatBtc, isValidTronAddress, tronAddressToHex } from './utils';
+import { estimateGasWithBuffer, formatBtc, isValidTronAddress, tronAddressToHex } from './utils';
 
 const RETRY_COUNT = 8; // Number of times to retry fetching transaction receipt after sending a transaction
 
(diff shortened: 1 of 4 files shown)


sdk/src/gateway/client.ts
```diff
@@ -63,6 +63,47 @@ function signerAccount(walletClient: WalletClient<Transport, ViemChain, Account>
...
+            to: tx.to,
+            data: tx.data,
+            value: tx.value,
+        });
+        return applyGasBuffer(estimate);
+    } catch {
+        return undefined;
+    }
+}
+
```
slavastartsev: this logic ideally needs to be stored in `utils/gas.ts` file

slavastartsev (approved):
looks good to me, thanks

PR #1104, opened 2026-07-03
docs(gateway): partner-questions updates — refunds, fees, onboarding, explorer links, editorial

zhixizhang:
## What

Gateway docs updates from the partner-questions ticket (BD-owned batch): https://app.notion.com/p/3793a8aad3b4804a800af71743abd7c9

- **Refunds rewrite** — documents the two real flows: BTC→X refunded automatically as USDT on Ethereum (price drift while BTC confirms); X→BTC manual claim after the claim delay (very rare). Adds at-a-glance table, clarifies BOB processes refunds and integrators don't need refund UI.
- **Fees** — typical-values table + worked example (illustrative, $100 BTC→USDT with 50 bps affiliate fee).
- **New Partner Onboarding page** (`/gateway/onboarding`, first Integration sidebar item) — API key flow, dashboard, where-to-find-what, FAQ, contact.
- **Order tracking** — documents `gateway-explorer.gobob.xyz/order/<ORDER_ID>` under Monitor Orders.
- **Partner dashboard** — documents `gateway-explorer.gobob.xyz/affiliate/<API_KEY>` under Monetization.
- **API key sections** — benefits list (rate limits, analytics dashboard, partner features) + link to onboarding page, version-neutral wording.
- **Editorial** — FAQ accordions dissolved into searchable headings; "BTC to X" phrasing in prose (API field names untouched); API Reference group added to Gateway tab; Alexei's gas-fee example nit.

## Correctness review — resolved
All [TO CONFIRM] items resolved per team feedback (Discord, 06–07/07): no refund-delay number documented (shortening in progress, bob-gateway#1622); refunds split by direction — BTC→X automatic USDT on Ethereum, X→BTC user-withdraws via BOB app activity section or refundTx; fees are quote-first with no published numbers; rate limits mentioned qualitatively only.

### DIFF
--- a/docs/gateway/api-reference/overview.mdx
+++ b/docs/gateway/api-reference/overview.mdx
@@ -37,7 +37,14 @@ V3 builds on V2 with multi-chain support and richer settlement data:
 
 ## Authentication
 
-API keys are **optional**. All V3 endpoints are reachable without authentication; a key unlocks higher rate limits and access to partner features. To request one, contact the BOB team.
+API keys are **optional** — the API is reachable without authentication. A key unlocks:
+
+- **Analytics dashboard** — your orders, volume, and affiliate earnings
+- **Higher rate limits** than keyless usage
+
+<Card title="Partner Onboarding" icon="rocket" href="/gateway/onboarding">
+  Get an API key and go live — full flow and contact details
+</Card>
 
 When you have a key, send it as a Bearer token on every V3 request:
 
--- a/docs/gateway/docs.json
+++ b/docs/gateway/docs.json
@@ -31,12 +31,19 @@
           {
             "group": "Integration",
             "pages": [
+              "gateway/onboarding",
               "gateway/integration",
               "gateway/migration",
               "gateway/agents",
               "gateway/strategies",
               "gateway/wallets",
               "gateway/faq"
+            ]
+          },
+          {
+            "group": "API Reference",
+            "pages": [
+              "api-reference/overview"
             ]
           }
         ]
--- a/docs/gateway/gateway/agents.mdx
+++ b/docs/gateway/gateway/agents.mdx
@@ -195,7 +195,7 @@ The [gateway-starter-bot](https://github.com/bob-collective/gateway-starter-bot)
 
 The bot executes a **"buy the dip"** round-trip:
 
-1. **Onramp** — swap BTC → USDT on Ethereum.
+1. **BTC to X** — swap BTC → USDT on Ethereum.
 2. **Watch** — poll the reverse quote (USDT → BTC) until BTC gets cheaper.
 3. **Buy the dip** — swap USDT → BTC, ending with more sats than it started with.
 
--- a/docs/gateway/gateway/faq.mdx
+++ b/docs/gateway/gateway/faq.mdx
@@ -6,72 +6,53 @@ icon: "circle-question"
 
 ## Frequently Asked Questions
 
-<AccordionGroup>
-
-<Accordion icon="gas-pump" title="Who pays the gas fees?">
+### Who pays the gas fees?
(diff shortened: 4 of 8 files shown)


zhixizhang:
@ducklings review

zhixizhang:
@ducklings review

gregdhill:
Nice cleanup — the refunds split and the fees rework read really well. Two things I'd like to adjust before merge, both about not pinning partners to internals:

**1. Don't have partners rely on `signedQuoteData` — `fees.mdx:9`**

The quote-first paragraph names the `signedQuoteData` field directly ("The quote is signed (`signedQuoteData`) and used to create the order…"). That's an internal implementation detail we don't want partners building against — if we rename or restructure it, their integrations break and it reads as a documented contract. Can we make the same point without naming the field? e.g.:

> "The quote you're shown is the quote that gets used to create the order, so the rate is the user's guaranteed minimum — no hidden costs and no surprises at settlement."

**2. Link supported chains to the dynamic routes page instead of hardcoding — `faq.mdx:47`**

The "What chains are supported?" answer hardcodes a chain list ("Bitcoin → Ethereum, Base, Arbitrum, Optimism (via LayerZero)"). It's already stale — it omits Tron/Solana and still says "via LayerZero", which contradicts the V3 messaging elsewhere in this same PR. Rather than maintain a list that drifts, let's point at [Supported Routes](/gateway/supported-routes), which fetches the live routes from the API. Something like:

> "Gateway supports Bitcoin and a growing set of EVM and non-EVM chains. See [Supported Routes](/gateway/supported-routes) for the current, live list."

(Same pattern applies to the hardcoded lists in `agents.mdx:19` and `strategies.mdx:45` — both pre-existing and out of scope here, but worth a follow-up so there's a single source of truth.)

nud3l:
@ducklings-app review

zhixizhang:
> Nice cleanup — the refunds split and the fees rework read really well. Two things I'd like to adjust before merge, both about not pinning partners to internals:
> 
> **1. Don't have partners rely on `signedQuoteData` — `fees.mdx:9`**
> 
> The quote-first paragraph names the `signedQuoteData` field directly ("The quote is signed (`signedQuoteData`) and used to create the order…"). That's an internal implementation detail we don't want partners building against — if we rename or restructure it, their integrations break and it reads as a documented contract. Can we make the same point without naming the field? e.g.:
> 
> > "The quote you're shown is the quote that gets used to create the order, so the rate is the user's guaranteed minimum — no hidden costs and no surprises at settlement."
> 
> **2. Link supported chains to the dynamic routes page instead of hardcoding — `faq.mdx:47`**
> 
> The "What chains are supported?" answer hardcodes a chain list ("Bitcoin → Ethereum, Base, Arbitrum, Optimism (via LayerZero)"). It's already stale — it omits Tron/Solana and still says "via LayerZero", which contradicts the V3 messaging elsewhere in this same PR. Rather than maintain a list that drifts, let's point at [Supported Routes](/gateway/supported-routes), which fetches the live routes from the API. Something like:
> 
> > "Gateway supports Bitcoin and a growing set of EVM and non-EVM chains. See [Supported Routes](/gateway/supported-routes) for the current, live list."
> 
> (Same pattern applies to the hardcoded lists in `agents.mdx:19` and `strategies.mdx:45` — both pre-existing and out of scope here, but worth a follow-up so there's a single source of truth.)

Thanks Greg — addressed both points in 11f4f4a2: dropped the signedQuoteData reference in fees.mdx and swapped the hardcoded chain list in faq.mdx for a link to the Supported Routes page.

gregdhill (approved):
Docs-only PR. Verified all internal links/anchors resolve, no stray MDX tags, worked-example math checks out, and payout currency is consistent with the USDT-on-Ethereum invariant. LGTM.

PR #1105, opened 2026-07-04
docs: add NodeFlare to node providers

BabyScope:
Adds NodeFlare to the node providers page, same format as existing entries.

NodeFlare serves BOB Mainnet (60808) from its own bare-metal nodes in two regions (Asia + US-East) with automatic failover — free public endpoint `https://rpc.nodeflare.app/bob/public` (no API key) plus a free keyed tier including `eth_getLogs` and debug/trace. Live and verifiable via `eth_chainId` → `0xed88`.


BabyScope:
Superseding from our project org — resubmitted from the NodeFlare org account.

PR #1113, opened 2026-07-13
feat(sdk): register HyperEVM (chainId 999) in supportedChainsMapping

nud3l:
## Problem

`gateway-cli` resolves chain metadata (chain id, native token, token decimals, multicall3) through the SDK's `supportedChainsMapping`. It lacked **HyperEVM (chainId 999)**, so BTC↔HyperEVM swaps couldn't resolve tokens/balances (`Chain id for "hyperliquid"/"hyperevm" not found`).

## Change

- Add `hyperevm` and `hyperliquid` keys to `supportedChainsMapping`, both → viem's `hyperEvm` (id 999). `hyperevm` is canonical; `hyperliquid` is kept as an alias because staging currently returns that name (bob-gateway#1656 canonicalizes it to `hyperevm`; merged, not yet deployed).
- Bump `viem` `^2.33.2` → **`2.49.0`**:
  - 2.49.0 is the first viem release that includes the **`multicall3`** contract (`0xcA11…CA11`) in its `hyperEvm` chain definition. Without it, multicall reads (e.g. gateway-cli `balance`) fail with `Chain "HyperEVM" does not support contract "multicall3"` — multicall3 is in fact deployed on HyperEVM.
  - Pinned below 2.55: viem ≥2.55 pulls a newer `ox` whose `.ts` sources fail the SDK's `tsc` build. 2.49.0 keeps `ox` 0.14.20 (clean build).

## Verification

- `npm run build` (tsc) ✅
- `getChainConfig('hyperevm')` → id 999, `contracts.multicall3` = `0xcA11…CA11` ✅
- `getChainConfig('hyperliquid')` → id 999 ✅ ; `getChainConfig('ethereum')` → 1 ✅
- Self-review (ducklings methodology): alias correctly absorbed by all consumers; no correctness findings.

## Follow-ups (separate gateway-cli PR)
Discovered while driving BTC↔HyperEVM swaps through the CLI:
- **Price oracle** can't price `USD₮0` (Coinbase/Binance by-symbol 404) → use the tokenlist `coingeckoId` (already emitted on tokenlist `main`; needs the CLI's tokenlist pin bumped + a CoinGecko-by-id source).
- **Chain-name coupling**: the CLI's route-derived token index is keyed by the API's chain string, so `:hyperevm` vs `:hyperliquid` must match what the API returns. Add alias normalization so both resolve across the API rename.


nud3l:
Note: viem is pinned exactly to `2.49.0` here because the SDK's classic `node10` module resolution can't build against the newer `ox` that viem >2.49 pulls. Modernizing the SDK's module resolution (to match the UI, which is on viem 2.53.1) is tracked in #1115 — once that lands, viem can be un-pinned to align with the UI.

PR #1114, opened 2026-07-13
feat(gateway-cli): HyperEVM support — coingeckoId pricing + chain aliasing

nud3l:
Follow-ups to bob#1113 (SDK) / bob-gateway#1656 (API), fixing CLI-level gaps found while running BTC↔HyperEVM swaps through `gateway-cli`.

## 1 — Price oracle can't price USD₮0 (prod bug)
`--amount NUSD` converts USD→token via `fetchPrice`, which queried Binance/Coinbase **by spot symbol**. Neither lists `USD₮0`, so any USD-denominated HyperEVM swap threw `PriceOracleError … Coinbase HTTP 404` — and the bot uses `--amount 1500USD`/`50USD`.

Add a **CoinGecko-by-id** source, tried first when the tokenlist provides a `coingeckoId`, threaded `getTokenMetadata → ResolvedAsset → parseAmount → fetchPrice`. Falls back to the existing exchange sources.

## 3 — Chain must be referred to by the API's exact name
The token index was keyed by the gateway's raw chain string, so `:hyperevm` failed while `:hyperliquid` worked (and that flips once the API canonicalizes to `hyperevm`). Key the index by **chainId** instead — any of a chain's names (both `hyperevm`/`hyperliquid` → 999 via the SDK) resolve — while the resolved asset keeps the **route's own chain name** for the outbound API call.

> Requires a `@gobob/bob-sdk` that maps both names → 999 (bob#1113) for the alias to activate at runtime. The CLI logic is version-agnostic and unit-tested with both names.

## 5 — Build was broken (stale SDK in lockfile)
The source already calls the 9-arg `createBitcoinPsbt` / 8-arg `estimateTxFee` (`utxoSelectionStrategy`) that `@gobob/bob-sdk` 5.8.x defines, but the lockfile pinned 5.7.1 (one fewer arg), so `tsc` failed. Refreshed the lock to 5.8.0 — no source change. Also bumped the `@gobob/tokenlist` pin so the emitted list carries `extensions.coingeckoId`.

## Verification
- `pnpm build` ✅ · `pnpm test` ✅ (106 tests; +5 new covering CoinGecko pricing and hyperevm↔hyperliquid resolution)
- Self-review (ducklings methodology): no correctness findings; two low-impact P3 notes (last-write-wins if the API ever returns two names for one chainId; native-token coingeckoId not surfaced).

## Follow-up
Once bob#1113 publishes a `@gobob/bob-sdk` with HyperEVM, bump the CLI's dep to it so the alias resolves at runtime, then cut a `@gobob/gateway-cli` release.

### DIFF
--- a/gateway-cli/package.json
+++ b/gateway-cli/package.json
@@ -21,7 +21,7 @@
   },
   "dependencies": {
     "@gobob/bob-sdk": "^5.8.0",
-    "@gobob/tokenlist": "github:bob-collective/tokenlist#5ee721890cb91657f640440a05698c357a080c87",
+    "@gobob/tokenlist": "github:bob-collective/tokenlist#5765810a897f2663f098939806d9f4285b9e8027",
     "commander": "^13.1.0",
     "p-retry": "^7.1.1",
     "viem": "^2.31.3",
--- a/gateway-cli/pnpm-lock.yaml
+++ b/gateway-cli/pnpm-lock.yaml
@@ -16,8 +16,8 @@ importers:
         specifier: ^5.8.0
         version: 5.8.0(@tanstack/react-query@5.91.0(react@19.2.4))(react-dom@19.2.4(react@19.2.4))(react@19.2.4)(typescript@5.9.3)(zod@4.3.6)
       '@gobob/tokenlist':
-        specifier: github:bob-collective/tokenlist#5ee721890cb91657f640440a05698c357a080c87
-        version: https://codeload.github.com/bob-collective/tokenlist/tar.gz/5ee721890cb91657f640440a05698c357a080c87
+        specifier: github:bob-collective/tokenlist#5765810a897f2663f098939806d9f4285b9e8027
+        version: https://codeload.github.com/bob-collective/tokenlist/tar.gz/5765810a897f2663f098939806d9f4285b9e8027(typescript@5.9.3)(zod@4.3.6)
       commander:
         specifier: ^13.1.0
         version: 13.1.0
@@ -258,8 +258,8 @@ packages:
       react: '>=18'
       react-dom: '>=18'
 
-  '@gobob/tokenlist@https://codeload.github.com/bob-collective/tokenlist/tar.gz/5ee721890cb91657f640440a05698c357a080c87':
-    resolution: {gitHosted: true, tarball: https://codeload.github.com/bob-collective/tokenlist/tar.gz/5ee721890cb91657f640440a05698c357a080c87}
+  '@gobob/tokenlist@https://codeload.github.com/bob-collective/tokenlist/tar.gz/5765810a897f2663f098939806d9f4285b9e8027':
+    resolution: {gitHosted: true, integrity: sha512-yS4vAV6+luoSJkLQglKT+/lCr+EfLpXV4CMRfeSkJOWT6C5pudJtGlGvO9dS5Km49+O5ZNK2573fTcDmgx/jvw==, tarball: https://codeload.github.com/bob-collective/tokenlist/tar.gz/5765810a897f2663f098939806d9f4285b9e8027}
     version: 1.0.0
 
   '@jridgewell/sourcemap-codec@1.5.5':
@@ -1522,7 +1522,15 @@ snapshots:
       - utf-8-validate
       - zod
 
-  '@gobob/tokenlist@https://codeload.github.com/bob-collective/tokenlist/tar.gz/5ee721890cb91657f640440a05698c357a080c87': {}
+  '@gobob/tokenlist@https://codeload.github.com/bob-collective/tokenlist/tar.gz/5765810a897f2663f098939806d9f4285b9e8027(typescript@5.9.3)(zod@4.3.6)':
+    dependencies:
+      '@scure/base': 1.2.6
+      viem: 2.47.5(typescript@5.9.3)(zod@4.3.6)
+    transitivePeerDependencies:
+      - bufferutil
+      - typescript
+      - utf-8-validate
+      - zod
 
   '@jridgewell/sourcemap-codec@1.5.5': {}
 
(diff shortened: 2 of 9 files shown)


nud3l:
@ducklings review

nud3l:
Addressed the review in the latest commit: added an overflow guard in `parseAmount` (rejects USD→token amounts ≥1e21 instead of crashing on `toFixed` exponential notation), and threaded `coingeckoId` through `resolveSendAsset` so the `send` path prices USD₮0 via CoinGecko too. @ducklings review

nud3l:
Addressed the review leads:

**1. Price oracle (429 masking + serial latency) — fixed** (latest commit). `fetchPrice` now queries CoinGecko + Binance + Coinbase **in parallel** and returns the first success in preference order (CoinGecko → Binance → Coinbase), so a price comes back whenever *any* source responds. When every source fails and a `coingeckoId` was queried, the error now surfaces the CoinGecko failure (e.g. the 429) instead of a misleading exchange-404 "unpriceable". Added tests for both.

**2. `coingeckoId` chain-accuracy — intended / acceptable.** The canonical fallback is a best-effort **asset-level** id used only for approximate USD→amount conversion; variants of the same asset price ~identically, and USD₮0 (the actual case here) is `uniform` so it always uses the canonical entry anyway. Documented inline in `getTokenMetadata`.

**3. `chainKey` needing both names in `CHAIN_IDS` — not applicable.** The new gateway environment only exposes **`hyperevm`** as a route (it won't return `hyperliquid`), so the token index is keyed by `hyperevm`'s chainId and the main path only needs `hyperevm` in `CHAIN_IDS` (provided by #1113). The `hyperliquid` alias is just extra tolerance for user input via the SDK — if it were ever dropped, only the `:hyperliquid` convenience alias would stop resolving, no real route breaks.

@ducklings review

nud3l:
@ducklings-app review

nud3l:
@ducklings-app review

nud3l:
Resolved the three review threads:

- **`send.ts:42` (P1)** — **Fixed.** `resolveSendAsset` now reads `extensions.coingeckoId` (via the `Meta`/tokenlist-index change) and includes it on the resolved asset, so a USD-denominated `send` of USD₮0 prices via CoinGecko instead of falling through to the exchange 404.
- **`input-resolver.ts:148` (P1)** — **Not a live issue.** With #1113 merged, `supportedChainsMapping` registers **both** `hyperevm` and `hyperliquid` → 999, and the new gateway environment only returns `hyperevm` as a route name — so `token.chain` always resolves through the SDK; the gateway/SDK names never diverge in a way that throws.
- **`input-resolver.ts:41` (P3)** — Same root: #1113 is merged, so the **real** `supportedChainsMapping` now contains both names (no longer mock-only), and the env returns only `hyperevm`.

nud3l:
Closed the P2 in code: `getTokenMetadata` no longer borrows the canonical entry's `coingeckoId` for a chain-specific token — it returns the chain-specific entry as-is, so a price is only ever resolved against the exact token. Uniform tokens (incl. USD₮0) are unchanged. @ducklings-app review

nud3l:
@ducklings-app review

nud3l:
Addressed the re-review:

- **`price-oracle.ts:102` (P2, latency) — fixed.** The exchange lookups now fire in parallel from the start while CoinGecko is awaited with a short 2.5s timeout; a CoinGecko success wins, otherwise the already-in-flight exchange result is used — so a slow/rate-limited CoinGecko no longer caps command latency.
- **`evm.ts:75` (P2, uniform coingeckoId) — fixed.** `coingeckoId` is now part of the `uniform` equality check, so same-address entries that differ only in `coingeckoId` are treated as chain-specific and never borrow another entry's id.
- **`asset-resolver.ts:87` (P3, native-token coingeckoId on send) — intended/accepted.** Native tokens (ETH, HYPE, …) resolve their price by symbol; the tokenlist's native-currency id isn't threaded through `getNativeToken`. Real BTC/ETH sends are exchange-listed and price fine; only an exotic native token unlisted on both exchanges would hit this, which is out of scope for this PR (swap-path HyperEVM pricing). Tracked as a known limitation.

@ducklings-app review

nud3l:
Flipped the oracle to **exchange-first** per discussion: Binance/Coinbase are queried first and preferred (authoritative for listed tokens); CoinGecko-by-id is consulted only when both exchanges fail (the exchange-unlisted case, e.g. USD₮0). This closes both the P1 (a wrong-but-positive CoinGecko value can no longer scale a swap amount when a real exchange price exists) and the P3 (exchange-listed tokens no longer wait on CoinGecko). USD₮0 still prices (exchanges 404 → CoinGecko). Added a test asserting the exchange price wins over a deliberately-wrong CoinGecko value. @ducklings-app review

nud3l:
@ducklings-app review

nud3l:
Simplified to end the review loop: **dropped the chainId-alias machinery** and reverted the token index to plain **chain-name keying** (the original approach), keeping the coingeckoId threading + the exchange-first oracle.

Rationale: the gateway env exposes HyperEVM under a single canonical name (`hyperevm`) — there's no `hyperliquid`→`hyperevm` rename transition to bridge — so the aliasing was solving a non-problem and was the root of the recurring edge-case findings (dedup-key mismatch, "both names must be in `CHAIN_IDS`", dual-name last-write-wins). Name-keying makes those structurally impossible.

**Verified against live staging** (which now returns `hyperevm`, so bob-gateway#1656 is deployed):
```
quote --src <USD₮0>:hyperevm --dst BTC --amount 47USD
→ srcAmount 47040314 (≈47.04 USD₮0), dstAmount 74166 sats
```
That exercises the full path — name resolution + the USD-denominated CoinGecko pricing that was previously broken — and it prices correctly (~$0.999) even with the currently-published SDK 5.8.0 (USD₮0 is uniform, so it doesn't need `CHAIN_IDS`).

All review threads resolved. Not auto-re-triggering ducklings, given the runs have been partial/degraded and the finding-generating machinery is now gone — happy to run one more if you'd like a clean confirmation.

PR #1125, opened 2026-07-16
fix(sdk): disambiguate GetMaxSpendableV2Request so the client compiles

nud3l:
## Why

`@gobob/bob-sdk` does not build on `master` — `tsc` fails with:

```
src/gateway/generated-client/apis/index.ts(5,1): error TS2308: Module './V2Api' has already exported a member named 'GetMaxSpendableV2Request'.
```

The generated client predates **bob-gateway#1658**, which gave v3 `get-max-spendable` its own operationId. Because the client was generated from the older spec, both `V2Api` and `V3Api` declare `GetMaxSpendableV2Request`, and the star re-exports in `apis/index.ts` make that ambiguous.

This blocks the SDK npm publish workflow (it runs `pnpm build`), which in turn blocks the HyperEVM release chain: the published SDK has no `hyperevm` in `supportedChainsMapping`, so the gateway-cli cannot resolve the chain (`getChainConfig("hyperevm")` throws), and a CLI release against it would ship broken HyperEVM support.

## What

One line: an explicit type re-export of `GetMaxSpendableV2Request` from `V2Api`, which takes precedence over the star exports and picks V2's copy (the two are structurally identical).

## Why not regenerate

A full `pnpm codegen` is the proper fix and gives v3 its own type — but it also pulls in unrelated generated-client changes from **bob-gateway#1609** (bungee v1→v3), which don't belong in a HyperEVM release. Kept surgical here. A future routine `pnpm codegen` will regenerate the client from the current spec and remove this stopgap; not separately tracked, to avoid nudging a regen into the in-flight bungee-v3 work.

## Verification

- `pnpm build` passes
- `pnpm test` passes (77 passed, 11 pre-existing skips)
- HyperEVM (`supportedChainsMapping`) present in the built output

### DIFF
--- a/sdk/src/gateway/generated-client/apis/index.ts
+++ b/sdk/src/gateway/generated-client/apis/index.ts
@@ -3,7 +3,7 @@
 export * from './V1Api';
 export * from './V2Api';
 export * from './V3Api';
-// STOPGAP (see follow-up issue): the spec already gives v3 its own operationId
+// STOPGAP: the spec already gives v3 its own operationId
 // (bob-gateway#1658), but this client was generated before that landed, so both V2Api and
 // V3Api still declare `GetMaxSpendableV2Request` — an ambiguous re-export (TS2308) that
 // breaks `tsc`. This explicit re-export wins over the star exports and picks V2's copy (the
--- a/sdk/src/gateway/generated-client/apis/index.ts
+++ b/sdk/src/gateway/generated-client/apis/index.ts
@@ -3,3 +3,10 @@
 export * from './V1Api';
 export * from './V2Api';
 export * from './V3Api';
+// STOPGAP (see follow-up issue): the spec already gives v3 its own operationId
+// (bob-gateway#1658), but this client was generated before that landed, so both V2Api and
+// V3Api still declare `GetMaxSpendableV2Request` — an ambiguous re-export (TS2308) that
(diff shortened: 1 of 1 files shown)


slavastartsev (approved):

sdk/src/gateway/generated-client/apis/index.ts
```diff
@@ -3,3 +3,10 @@
 export * from './V1Api';
 export * from './V2Api';
 export * from './V3Api';
+// STOPGAP: the spec already gives v3 its own operationId
+// (bob-gateway#1658), but this client was generated before that landed, so both V2Api and
+// V3Api still declare `GetMaxSpendableV2Request` — an ambiguous re-export (TS2308) that
+// breaks `tsc`. This explicit re-export wins over the star exports and picks V2's copy (the
+// two are structurally identical). Delete this line the next time `pnpm codegen` regenerates
+// the client from the current spec.
+export type { GetMaxSpendableV2Request } from './V2Api';
```
slavastartsev:
```suggestion
export type { type GetMaxSpendableV2Request as GetMaxSpendableV3Request } from './V2Api';
```

or you can apply this

slavastartsev: also, don't forget to bump the version in package.json

PR #1140, opened 2026-08-06
feat(sdk): add MempoolClient.getAddressInfo for address funding totals

danielsimao:
Adds `MempoolClient.getAddressInfo`, plus `totalReceived` / `confirmedReceived` helpers.

## Why

The swap tracking modal in `bob-collective/ui` cannot tell an underpaid Gateway deposit from a correct one, so a user who sends too little sits on "Waiting for deposit" with no explanation until the order expires.

Answering that needs one number: how much has this deposit address received.

## Why this shape

The Gateway solver already answers it, and the UI has to agree with it — if the two compute the number differently, a divergence means telling a user they are short while the Gateway is happy, or the reverse.

`bob-gateway crates/solver/src/driver.rs:51`:

```rust
let total_received = address_info.total_received();
if total_received > 0 && total_received < expected_balance {
    tracing::warn!(..., "Onramp order received insufficient amount");
}
if total_received >= expected_balance { /* proceed */ }
```

`crates/bitcoin-api/src/esplora_client.rs:117`:

```rust
pub fn total_received(&self) -> u64 {
    self.confirmed_received() + self.mempool_stats.funded_txo_sum
}
pub fn confirmed_received(&self) -> u64 {
    self.chain_stats.funded_txo_sum
}
```

So `getAddressInfo` returns that same response, and the two helpers apply the same arithmetic — a consumer cannot derive it differently.

Funding totals are also the right primitive rather than a balance: `funded_txo_sum` ignores anything since spent, so the number survives the deposit address being swept. A balance would read zero after a sweep and call a paid deposit underpaid.

Paging disappears along with the transaction list, and with it the page-size constant, the cursor, and the endpoint-shape question.

## Test plan

- [x] `vitest run test/mempool.test.ts` — 8 passed
- [x] `tsc --noEmit` clean
- [x] Rebased on `master` (`a8061d8b`)
- [ ] Release, so the consuming PR can bump from 5.8.1

The fixture is a swept address — 50k confirmed and since spent, 8k still pending — so a balance-based implementation returns 8,000 and fails the assertion of 58,000.

## History

Earlier revisions added `getAddressTxs` and then `getAddressChainTxs`, summing transaction outputs. Neither shipped. Tracing the consumer showed the total was already available in one call, computed by the system of record, without pagination.

## Downstream

Consumed by an underpaid-deposit note in the swap tracking modal (`bob-collective/ui`, follows this release). Nothing else depends on it.

### DIFF
--- a/sdk/src/mempool.ts
+++ b/sdk/src/mempool.ts
@@ -270,31 +270,12 @@ export class MempoolClient {
     }
 
     /**
-     * Get transactions for a Bitcoin address, confirmed and unconfirmed alike.
-     *
-     * Returns a single page: up to 50 mempool transactions plus the newest
-     * {@link MempoolClient.ADDRESS_TXS_PAGE_SIZE} confirmed ones. Because the
-     * two kinds share one array, its length says nothing about whether more
-     * confirmed history remains — use {@link getAddressChainTxs} to page that.
-     *
-     * @param {string} address - The Bitcoin address to check.
-     * @returns {Promise<MempoolTxInfo[]>} Newest transactions, mempool included.
-     *
-     * @example
-     * const mempoolClient = new MempoolClient();
-     * const txs = await mempoolClient.getAddressTxs('bc1q...');
-     */
-    async getAddressTxs(address: string): Promise<MempoolTxInfo[]> {
-        return this.getJson<MempoolTxInfo[]>(`${this.basePath}/address/${address}/txs`);
-    }
-
-    /**
      * Get confirmed transactions for a Bitcoin address, newest first.
      *
-     * Returns {@link MempoolClient.ADDRESS_TXS_PAGE_SIZE} per page. Unlike
-     * {@link getAddressTxs} the page holds nothing but confirmed transactions,
-     * so a short page means the history is exhausted. Totalling received value
-     * from one call under-counts any address with a longer history.
+     * Returns {@link MempoolClient.ADDRESS_TXS_PAGE_SIZE} per page, confirmed
+     * transactions only, so a short page means the history is exhausted.
+     * Totalling received value from one call under-counts any address with a
+     * longer history. Pair with {@link getAddressMempoolTxs} for a full picture.
      *
      * @param {string} address - The Bitcoin address to check.
      * @param {string} [lastSeenTxid] - Continue after this txid, from a prior page.
--- a/sdk/src/mempool.ts
+++ b/sdk/src/mempool.ts
@@ -98,10 +98,32 @@ export type BlockDetails = {
     difficulty: number;
 };
 
-export class MempoolClient {
-    /** Confirmed transactions per page from {@link MempoolClient.getAddressChainTxs}. */
-    static readonly ADDRESS_TXS_PAGE_SIZE = 25;
+export type AddressStats = {
+    tx_count: number;
+    funded_txo_count: number;
+    /** Total ever received, ignoring anything since spent. */
+    funded_txo_sum: number;
+    spent_txo_count: number;
+    spent_txo_sum: number;
+};
 
+export type AddressDetails = {
+    address: string;
+    chain_stats: AddressStats;
+    mempool_stats: AddressStats;
+};
+
+/** Confirmed funding received by an address. Survives the outputs being spent. */
+export function confirmedReceived(info: AddressDetails): number {
+    return info.chain_stats.funded_txo_sum;
+}
+
+/** Funding received by an address, confirmed and unconfirmed. */
+export function totalReceived(info: AddressDetails): number {
+    return confirmedReceived(info) + info.mempool_stats.funded_txo_sum;
+}
+
+export class MempoolClient {
     private basePath: string;
 
     /**
@@ -270,31 +292,17 @@ export class MempoolClient {
     }
 
     /**
-     * Get confirmed transactions for a Bitcoin address, newest first.
-     *
-     * Returns {@link MempoolClient.ADDRESS_TXS_PAGE_SIZE} per page, confirmed
-     * transactions only, so a short page means the history is exhausted.
-     * Totalling received value from one call under-counts any address with a
-     * longer history. Pair with {@link getAddressMempoolTxs} for a full picture.
+     * Get funding and spending totals for a Bitcoin address.
      *
      * @param {string} address - The Bitcoin address to check.
-     * @param {string} [lastSeenTxid] - Continue after this txid, from a prior page.
-     * @returns {Promise<MempoolTxInfo[]>} One page of confirmed transactions.
+     * @returns {Promise<AddressDetails>} Confirmed and mempool statistics.
      *
      * @example
      * const mempoolClient = new MempoolClient();
-     * const txs: MempoolTxInfo[] = [];
-     * let page = await mempoolClient.getAddressChainTxs('bc1q...');
-     * while (page.length === MempoolClient.ADDRESS_TXS_PAGE_SIZE) {
-     *     txs.push(...page);
-     *     page = await mempoolClient.getAddressChainTxs('bc1q...', page[page.length - 1].txid);
-     * }
-     * txs.push(...page);
+     * const received = totalReceived(await mempoolClient.getAddressInfo('bc1q...'));
      */
-    async getAddressChainTxs(address: string, lastSeenTxid?: string): Promise<MempoolTxInfo[]> {
-        const cursor = lastSeenTxid ? `/${lastSeenTxid}` : '';
-
-        return this.getJson<MempoolTxInfo[]>(`${this.basePath}/address/${address}/txs/chain${cursor}`);
+    async getAddressInfo(address: string): Promise<AddressDetails> {
+        return this.getJson<AddressDetails>(`${this.basePath}/address/${address}`);
     }
 
     /**
--- a/sdk/src/mempool.ts
+++ b/sdk/src/mempool.ts
@@ -99,7 +99,7 @@ export type BlockDetails = {
 };
 
 export class MempoolClient {
-    /** Confirmed transactions per page from {@link MempoolClient.getAddressTxs}. */
+    /** Confirmed transactions per page from {@link MempoolClient.getAddressChainTxs}. */
     static readonly ADDRESS_TXS_PAGE_SIZE = 25;
 
     private basePath: string;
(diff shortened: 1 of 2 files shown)


danielsimao:
@ducklings-app review

Small addition — one method plus tests — but two things are worth the scrutiny.

## The endpoint pair

```
/address/{address}/txs/mempool   ← existing getAddressMempoolTxs, unconfirmed only
/address/{address}/txs           ← new getAddressTxs, confirmed included
```

The mempool route is a **suffix** of the full one. Any mock or router matching with `includes()` resolves both to the same handler, so a test written that way passes while the code calls the wrong endpoint — the new method would ship as a silent duplicate of the existing one. The added mocks match on the full path with the mempool branch ordered first, and one test asserts the two calls return different results.

**Worth checking:** whether that ordering dependency is fragile enough to deserve a different mocking approach, and whether anything else in the client has the same prefix-collision shape.

## Semantics

The reason for the method is that `getAddressMempoolTxs` cannot answer "how much has this address received": a deposit leaves that list the moment it confirms, and once the address is swept its UTXOs go too. `/txs` keeps spent outputs in history, which is what makes totalling possible.

**Worth checking:** the upstream endpoint is paginated and returns most-recent-first with a server-side cap. The doc comment says so, but the method does not expose pagination. If a caller totalling deposits could realistically exceed one page, that is a correctness trap rather than a nicety — I would rather know now than after the consuming PR is written.

## Consumer

`bob-collective/ui` needs this to tell an underpaid Gateway deposit from a correct one; today a user who underpays sits on "Waiting for deposit" with no explanation. That work is blocked on this releasing. Our internal gateway-admin explorer already reads this endpoint directly for the same purpose — this brings it into the SDK instead of the app bypassing the client with a raw `fetch`.

No behaviour change to any existing method.

danielsimao:
Good catch — P1 accepted and fixed in 84432601.

I checked the mechanism before implementing, because I thought pagination here was a path segment (`/txs/chain/:last_seen_txid`). That is Blockstream Esplora's convention; mempool.space uses the `after_txid` query parameter, as the finding says. Your fix was right and my doubt was not.

## What changed

```ts
async getAddressTxs(address: string, afterTxid?: string): Promise<MempoolTxInfo[]>
```

- `afterTxid` appends `?after_txid=<value>`.
- `MempoolClient.ADDRESS_TXS_PAGE_SIZE` (25) is exported, so a caller can tell a full page from a final one — without it there is no way to detect truncation.
- The doc comment now carries the loop instead of leaving it implied, and states the cap explicitly.

## Why the cursor rather than an internal loop

You offered both. Paging stays the caller's decision: an internal loop would issue unbounded sequential requests for a heavily used address, and no caller of this method wants that implicitly. The consumer that prompted this reads a single-use Gateway deposit address, which will not span a page in practice — but the SDK should not encode that assumption, which is precisely your point.

## On the severity

Worth recording why this mattered more than a typical truncation bug. The consuming work sums received value to spot an underpaid deposit. Under-counting there does not merely lose data — it reports a **full payment as underpaid**, telling a user their money is short when it is not. That is the same failure the downstream PR exists to remove, so shipping it in the foundation would have been self-defeating.

Two tests added: paging past the first page returns different transactions, and the page-size constant is exposed at 25.

danielsimao:
@ducklings-app review

Two commits since your review.

**`84432601` — the P1.** `getAddressTxs` now takes an optional `afterTxid` and appends `?after_txid=`. `MempoolClient.ADDRESS_TXS_PAGE_SIZE` (25) is exported so a caller can tell a full page from a final one; without it there is no truncation signal to act on.

I verified the mechanism before implementing — I had thought this endpoint paged by path segment (`/txs/chain/:last_seen_txid`), which is Blockstream Esplora's convention, not mempool.space's. Your `after_txid` was right.

Paging is left to the caller rather than looped internally, because an internal loop would issue unbounded sequential requests for a heavily used address. Worth challenging if you disagree — you offered both.

**`c3dccf45` — trim.** The doc comment had grown to 31 lines against a 10-line neighbour, and two of the four tests asserted nothing about the code: one compared the page-size constant to its own literal, the other checked that a fixture contained a confirmed transaction when the method only builds a URL.

**Please check the trim did not cut something load-bearing.** The pagination caveat was deliberately kept; everything removed was either commit-history material or restating the line above it. If the remaining two tests leave a gap — particularly around the `/txs` vs `/txs/mempool` prefix collision — that is the thing to flag.

danielsimao:
Fixed in `19408d52` — but the two P2s did the work here, not the P1.

## P1 — rationale is wrong, conclusion accidentally right

The finding says mempool.space "ignores unrecognized query parameters" on `/address/{address}/txs`. It does not. Their REST docs for that endpoint:

> "Returns up to 50 mempool transactions plus the first 25 confirmed transactions. You can request more confirmed transactions using an `after_txid` query parameter."

So `?after_txid=` was documented and working. Worth flagging because this reverses your first review, which told me to use `after_txid` — I had originally reached for the path segment and changed it on your advice. I checked the docs both times rather than following whichever review was most recent.

## P2 — correct, and decisive

Both P2s land. `/txs` returns mempool **and** confirmed in one array, so its length measures something other than confirmed-page capacity. Your example is exact: 30 unconfirmed + 3 confirmed returns 33 entries, never falls below 25, and the documented loop never terminates while making no progress.

Counting only confirmed entries patches it. Using `/address/{address}/txs/chain` removes it — that endpoint returns confirmed transactions and nothing else, 25 per page, so a short page means exhausted and the constant finally measures the same thing as the array.

## What shipped

- `getAddressTxs(address)` — uncursored again, current picture including mempool.
- `getAddressChainTxs(address, lastSeenTxid?)` — pages confirmed history by path segment.

The path form is also what a plain Esplora instance accepts, which matters here: `REGTEST_MEMPOOL_BASE_PATH` and `REGTEST_ESPLORA_BASE_PATH` are both `http://localhost:3003`, so this client already talks to Esplora in regtest, where `after_txid` is a mempool.space extension that does not exist.

The mocks now key on real API paths rather than the URL this client happens to build — your point about the mock being built around the SDK's own construction was fair, and it applied to my paging test as written. Appending a query-string cursor now falls through to the uncursored handler and fails.

## The encodeURIComponent lead

Not addressed, deliberately. `getTxInfo`, `getBlock`, `getAddressMempoolTxs` and the rest all interpolate unencoded, so encoding only the two new methods would leave the client inconsistent without making it safe. Valid addresses and txids contain no URL-significant characters, and malformed input 404s either way. Worth its own change across the client if you think it earns one.

danielsimao:
@ducklings-app review

`19408d52` restructures rather than patches the paging finding.

Your P2s were right that `/address/{address}/txs` mixes mempool and confirmed entries, so array length cannot signal exhaustion. Rather than count confirmed entries against the constant, the cursor moved to `/address/{address}/txs/chain`, which returns confirmed transactions only, 25 per page — so a short page means exhausted and the constant measures the same thing as the array.

- `getAddressTxs(address)` — uncursored, current picture including mempool.
- `getAddressChainTxs(address, lastSeenTxid?)` — confirmed history, paged by path segment.

Mocks now key on real API paths rather than the URL this client builds, so a query-string cursor falls through to the uncursored handler and fails the test. That was your point about the mock being written around the SDK's own construction, and it was fair.

**Two things to check, in priority order.**

1. **The split itself.** Two methods where there was one. If `getAddressChainTxs` should instead be the only address-history method, or the pair invites a caller to reach for the wrong one, that is worth saying now rather than after downstream code is written against it.

2. **Regtest compatibility.** `REGTEST_MEMPOOL_BASE_PATH` and `REGTEST_ESPLORA_BASE_PATH` are both `http://localhost:3003`, so this client already talks to a plain Esplora instance in regtest. The path-segment form works on both; `after_txid` is a mempool.space extension. I believe that makes the path form strictly safer here — please check that reasoning rather than take it.

**On the previous P1:** its stated rationale was that mempool.space ignores unrecognized query parameters on `/txs`. Their REST docs document `after_txid` for that endpoint, so that was not the reason the change was needed — the P2s were. Flagging so the same rationale is not reused; it also reversed the first review, which had told me to use `after_txid`.

**Not addressed:** the `encodeURIComponent` lead. Every existing method in this client interpolates unencoded, so encoding only the two new ones leaves it inconsistent without making it safe. Worth its own change across the client if you think it earns one.

danielsimao:
@ducklings-app review

`161a193b` replaces the method entirely. Your earlier findings are resolved by removal rather than repair, so this is worth re-reading from scratch rather than as a diff.

## What changed and why

Previous revisions added `getAddressTxs`, then `getAddressChainTxs`, and had the consumer sum transaction outputs to detect an underpaid Gateway deposit. Tracing the consumer showed that answers the question a second way, in parallel with the system that actually decides it.

`bob-gateway crates/solver/src/driver.rs:51` — the solver's decision:

```rust
let total_received = address_info.total_received();
if total_received >= expected_balance { /* proceed */ }
```

`crates/bitcoin-api/src/esplora_client.rs:117` — what that means:

```rust
pub fn total_received(&self) -> u64 {
    self.confirmed_received() + self.mempool_stats.funded_txo_sum
}
pub fn confirmed_received(&self) -> u64 {
    self.chain_stats.funded_txo_sum
}
```

So the PR now adds `getAddressInfo` returning that same response, with `totalReceived`/`confirmedReceived` applying the same arithmetic. If the UI computed it differently it could tell a user they are short while the Gateway is satisfied. Pagination, the page-size constant and the endpoint-shape question all disappear with the transaction list.

## What I most want challenged

**1. Is `total_received` really the only input to the sufficiency decision?** I read `update_txid` (`driver.rs:350-398`) as selecting *which* txid to record — it filters out transactions that spend from the deposit address and picks the highest-paying one — without feeding the amount check. If any other path gates on per-transaction amounts, the helpers are the wrong abstraction.

**2. Address reuse.** `funded_txo_sum` is cumulative for an address's lifetime, so a reused deposit address would over-count and call an underpaid deposit sufficient — the inverse failure, and worse. `bob-gateway migrations/2026-03-12-120000-0000_prevent_duplicate_deposit_address/up.sql` adds a `BEFORE INSERT` trigger raising `unique_violation` on a duplicate `deposit_address`, and its own comment says it does not cover pre-existing rows or `UPDATE`. I judged that acceptable because the consumer renders only for in-flight orders, and because a duplicate would mislead the solver identically. Check that reasoning.

**3. The helpers.** `totalReceived`/`confirmedReceived` exist so a caller cannot re-derive the formula wrongly. If exported free functions are the wrong shape for this SDK, or they belong on the client, say so.

**4. Swept addresses.** The test fixture is deliberately swept — 50k confirmed and since spent, 8k pending — so a balance-based implementation returns 8,000 instead of 58,000 and fails. Confirm that actually distinguishes the two, since it is the property the whole choice rests on.

## Not addressed

The `encodeURIComponent` lead from your earlier pass. Every method in this client interpolates unencoded, so encoding one leaves it inconsistent without making it safe. Worth its own change across the client if you think it earns one.

danielsimao:
P3 confirmed intended, and the doc half is fixed in `15364531`.

**The solver already makes the distinction you are asking about.** `crates/solver/src/driver.rs:63-66`:

```rust
if total_received >= expected_balance {
    let txid = update_txid(self, &driver.runner.bitcoin).await?;

    if address_info.confirmed_received() >= expected_balance {
        // execute on BOB
```

`total_received` gates recording a txid. `confirmed_received` gates the irreversible step. So no funding is credited before finality — the outer check only decides whether it is worth looking for a transaction at all.

Your point still stands for a *caller* reading the helper in isolation, so `totalReceived` now documents that the unconfirmed part is provisional, can go down, and that anything irreversible should gate on `confirmedReceived`.

For the record on the consumer: the UI uses `totalReceived` to tell a user their deposit is short. Counting a pending top-up there is the behaviour we want — it is a message, not a release of funds, and being told you are still short after you have already sent the difference would be worse than the alternative.

danielsimao:
@ducklings-app review

`196f3510` shrinks the public surface after approval, so this is worth a fresh look at the API rather than a diff against what you last saw.

## What changed

`confirmedReceived` is gone. It had no caller — only its own test and `totalReceived` using it internally — and `src/index.ts` re-exports the whole module, so shipping it would have made a later removal breaking. Its field access is inlined and its test removed.

`totalReceived` stays exported, and that asymmetry is deliberate. It is also one line, but it encodes the rule that matters: add confirmed and pending funding, never subtract spent. `EsploraClient.getBalance` on the same endpoint does subtract, and would read zero once the Gateway solver sweeps a deposit address — turning a fully paid order into an apparent shortfall. The helper exists to stop the next caller reaching for balance.

## Worth attacking

- **The remaining surface.** `getAddressInfo` returns raw `AddressDetails` rather than a computed figure, on the reasoning that `getBalance` already demonstrates what happens when this endpoint is wrapped in one interpretation. Is raw-stats-plus-one-helper the right shape, or should the interpretation live in the client?
- **The duplication I chose not to fix.** `MempoolClient.getAddressInfo` and `EsploraClient.getBalance` hit the same path and parse the same JSON on two classes with different base paths. I left it. Say if you think that is wrong.
- **`totalReceived` mirrors a backend definition I cannot verify from this repo.** The consumer compares it against an order amount to decide whether a deposit is short, and the whole design rests on the Gateway solver computing `chain_stats.funded_txo_sum + mempool_stats.funded_txo_sum` too. The Rust source is not here. If the solver differs, the consumer is wrong in a way no test in this PR would catch.
- **Unconfirmed funds can go down.** RBF replacement or eviction lowers `mempool_stats`, so this figure is not monotonic. The doc says gate irreversible work on `chain_stats` alone. Is that enough, or should the type make it harder to misuse?

## Consumer

The only production caller is bob-collective/ui#2717, which is still draft and blocked on this releasing.

slavastartsev (approved):

danielsimao:
Closing alongside bob-collective/ui#2717, its only consumer.

`MempoolClient.getAddressInfo` was added so the UI could total a deposit address and detect an underpaid onramp. That read belongs to the backend instead: the solver already computes the same figure (`crates/solver/src/driver.rs:53`) and validates it against additional esplora sources before acting on it (`driver.rs:79`), which a single client-side read cannot match. Asked for as a field on the order instead — bob-collective/bob-gateway#<TBD>.

Shipping this now would put a method on the published SDK with no consumer, and the barrel re-exports everything in the module, so removing it later would be breaking — the same reason `confirmedReceived` was dropped from this branch during review.

Reopenable if a genuine consumer appears.

danielsimao:
Backend ask filed as bob-collective/bob-gateway#1793 — replaces the `<TBD>` above.

PR #1146, opened 2026-08-14
fix(sdk): wait for approval to be visible before sending

slavastartsev:
## Problem

`executeQuote` approves, awaits the approve receipt, then immediately sends the swap tx. `waitForTransactionReceipt` can resolve **before** the approval is reflected in `latest` state — Base serves flashblock preconfirmations, and load-balanced RPCs can answer the follow-up `eth_call` / `eth_estimateGas` from a node that has not applied the block yet.

viem estimates gas during `sendTransaction`, so the estimate reverts and the swap tx is **never broadcast**, even though the approval succeeded on-chain:

```
Execution reverted with reason: ERC20: transfer amount exceeds allowance.
Request Arguments:
  from:   0xAF91558Ba2B1994530c9cfCcbda5AE9cD2b456bb
  to:     0x0000000000001fF3684f28c67538d4D072C22734
Details: execution reverted: ERC20: transfer amount exceeds allowance
Version: viem@2.49.0
```

Observed in the gateway-bot API test workflow, leg `USDC@base → BTC` ([run 31794363730](https://github.com/bob-collective/gateway-bot/actions/runs/31794363730/job/94748230262)):

| Time (UTC) | Base block | Fact |
|---|---|---|
| 11:01:21 | 49957967 | `allowance(bot, 0x…22734)` = **0** |
| 11:01:21.7 | — | CLI throws `ERC20: transfer amount exceeds allowance` |
| 11:01:23 | 49957968 | approve tx `0x9f40d9c4…` (nonce 208) mined → allowance = 24975524 |

The approved amount equals exactly what the swap calldata pulls (`0x17d18a4`), and the wallet nonce is `209` — nonce 208 (the approve) was the last one used, confirming the offramp tx was never sent. A `cast call --trace` of the calldata also confirms the puller and the approved spender are the same address, so this is purely a state-visibility race, not a wrong-spender or amount bug.

## Fix

Add `waitForAllowance` (`sdk/src/gateway/utils/allowance.ts`), which polls `allowance` until it covers the required amount, and call it after the approve receipt in the offramp, tokenSwap and `executeStrategy` paths.

- Checks before sleeping → common case costs one extra `eth_call` and no delay.
- Tolerates transient RPC read failures.
- On exhaustion (default 10 × 500ms) throws with the observed allowance, spender, required amount and approve tx hash. Failing there is preferable to a revert during the send, because nothing has been signed yet.

The USDT reset-to-0 step is left unchanged — same class of race, but not the observed failure.

## Notes

`estimateGasWithBuffer` does not protect against this: it swallows the revert and returns `undefined`, after which viem re-estimates during the send and throws the same error.

## Test plan

- `sdk/test/allowance.test.ts` (new): immediate return, polls through stale reads, survives a transient RPC error, throws with the observed allowance after N attempts.
- `sdk/test/gateway.test.ts`: `mockOftReadContract` now models post-approve state (`allowanceAfterApproval`, default `maxUint256`); two new offramp tests assert the re-check happens before `sendTransaction`, and that a permanently-stale allowance rejects without sending.
- `pnpm run test` → 83 passed | 11 skipped. `pnpm run lint` and `pnpm run build` clean.

### DIFF
--- a/gateway-cli/package.json
+++ b/gateway-cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@gobob/gateway-cli",
-  "version": "0.4.0",
+  "version": "0.4.1",
   "description": "CLI for bridging Bitcoin to/from EVM chains via BOB Gateway",
   "repository": {
     "type": "git",
@@ -20,7 +20,7 @@
     "cli:dev": "tsx bin/gateway-cli.ts"
   },
   "dependencies": {
-    "@gobob/bob-sdk": "^5.9.0",
+    "@gobob/bob-sdk": "^5.9.1",
     "@gobob/tokenlist": "github:bob-collective/tokenlist#5765810a897f2663f098939806d9f4285b9e8027",
     "commander": "^13.1.0",
     "p-retry": "^7.1.1",
--- a/gateway-cli/pnpm-lock.yaml
+++ b/gateway-cli/pnpm-lock.yaml
@@ -13,8 +13,8 @@ importers:
   .:
     dependencies:
       '@gobob/bob-sdk':
-        specifier: ^5.9.0
-        version: 5.9.0(@tanstack/react-query@5.91.0(react@19.2.4))(react-dom@19.2.4(react@19.2.4))(react@19.2.4)(typescript@5.9.3)(zod@4.3.6)
+        specifier: ^5.9.1
+        version: 5.9.1(@tanstack/react-query@5.91.0(react@19.2.4))(react-dom@19.2.4(react@19.2.4))(react@19.2.4)(typescript@5.9.3)(zod@4.3.6)
       '@gobob/tokenlist':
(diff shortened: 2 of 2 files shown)


sdk/src/gateway/client.ts
```diff
@@ -407,6 +407,13 @@ export class GatewayApiClient {
                 callback?.({ step: needsReset ? 2 : 1, type: ExecuteQuoteStepType.Approve, totalSteps });
                 const approveTxHash = await walletClient.writeContract(request);
                 await publicClient.waitForTransactionReceipt({ hash: approveTxHash, retryCount: RETRY_COUNT });
```
slavastartsev, line 409: also another option is to simply set `confirmations: 3` or higher number
@danielsimao thoughts?

danielsimao: I would not go above 3 confirmations. 3 should be fine here

danielsimao (approved):

PR #1149, opened 2026-08-17
feat(sdk): surface orderId in executeQuote callbacks and thrown errors

danielsimao:
## Problem

`executeQuote` creates the Gateway order first, then runs the remaining steps (BTC signing for onramp; approval + send for offramp/tokenSwap). If any of those later steps fail — most notably a wallet rejecting or erroring on the BTC signing step — the caller has no `orderId` available at the failure point. Errors thrown after order creation carry no order context, and the `ExecuteQuoteStep` callback payload never included one either. Consumers instrumenting these failures (e.g. error reporting) can't cross-reference against the Gateway's own order record, even though the order already exists server-side by that point.

## Change

- `ExecuteQuoteStep` (`sdk/src/gateway/types/execute.ts`) gains an optional `orderId?: string` field. Backward compatible — existing consumers that don't read it are unaffected.
- Added `ExecuteQuoteError` (extends `Error`, optional `orderId?: string`) for the throw path.
- In `executeQuote`, once `order` is created and its type is confirmed for each flow (onramp/offramp/tokenSwap), every subsequent `callback?.(...)` invocation now includes `orderId`.
- Each flow's remaining logic (after order creation) is wrapped in a `try/catch` that tags any thrown error with `orderId` via `attachOrderId` before rethrowing — mutating the error in place so its identity, message, and stack are preserved for existing callers, rather than wrapping it in a new error.

No behavioral changes to signing, approval, or send logic — this is purely additive metadata propagation.

## Test plan

- `pnpm run build` (tsc) — clean
- `pnpm run lint` (eslint) — clean
- `pnpm run format:check` (prettier) — clean
- `pnpm run test` (vitest) — 79 passed | 11 skipped
- Updated existing callback assertions in `test/gateway.test.ts` to expect `orderId`
- Added 3 new tests: onramp throw-after-order-creation carries `orderId`, offramp send-transaction failure carries `orderId`, tokenSwap send-transaction failure carries `orderId`

### DIFF
--- a/sdk/src/gateway/client.ts
+++ b/sdk/src/gateway/client.ts
@@ -71,7 +71,7 @@ async function simulateApproval<T>(orderId: string, simulate: () => Promise<{ re
     try {
         return (await simulate()).request;
     } catch (error) {
-        throw new ExecuteQuoteError(orderId, translateApprovalError(error));
+        throw new ExecuteQuoteError(orderId, { cause: translateApprovalError(error) });
     }
 }
 
@@ -322,14 +322,13 @@ export class GatewayApiClient {
                     });
                     if (!bitcoinTxHex) throw new Error('Failed to get signed transaction');
                 } catch (error) {
-                    throw new ExecuteQuoteError(orderId, error);
+                    throw new ExecuteQuoteError(orderId, { cause: error });
                 }
             } else if (btcSigner.signAllInputs) {
                 if (!order.onramp.psbtHex) {
-                    throw new ExecuteQuoteError(
-                        orderId,
-                        new Error('PSBT not available: sender address is required when using signAllInputs')
-                    );
+                    throw new ExecuteQuoteError(orderId, {
+                        message: 'PSBT not available: sender address is required when using signAllInputs',
+                    });
                 }
                 const psbtHex = order.onramp.psbtHex;
                 callback?.({
@@ -342,13 +341,12 @@ export class GatewayApiClient {
                     bitcoinTxHex = await btcSigner.signAllInputs(psbtHex);
                     if (!bitcoinTxHex) throw new Error('Failed to get signed transaction');
                 } catch (error) {
-                    throw new ExecuteQuoteError(orderId, error);
+                    throw new ExecuteQuoteError(orderId, { cause: error });
                 }
             } else {
-                throw new ExecuteQuoteError(
-                    orderId,
-                    new Error('btcSigner must implement either sendBitcoin or signAllInputs method')
-                );
+                throw new ExecuteQuoteError(orderId, {
+                    message: 'btcSigner must implement either sendBitcoin or signAllInputs method',
+                });
             }
 
             let tx: RegisterTxSuccess;
@@ -373,7 +371,7 @@ export class GatewayApiClient {
                     tx = response;
                 }
             } catch (error) {
-                throw new ExecuteQuoteError(orderId, error);
+                throw new ExecuteQuoteError(orderId, { cause: error });
             }
 
             if (typeof tx === 'string') {
@@ -430,7 +428,7 @@ export class GatewayApiClient {
                             args: [accountAddress, spenderAddress],
                         });
                     } catch (error) {
-                        throw new ExecuteQuoteError(orderId, error);
+                        throw new ExecuteQuoteError(orderId, { cause: error });
                     }
                 }
             }
@@ -466,7 +464,7 @@ export class GatewayApiClient {
                         const resetTxHash = await walletClient.writeContract(resetRequest);
                         await publicClient.waitForTransactionReceipt({ hash: resetTxHash, retryCount: RETRY_COUNT });
                     } catch (error) {
-                        throw new ExecuteQuoteError(orderId, translateApprovalError(error));
+                        throw new ExecuteQuoteError(orderId, { cause: translateApprovalError(error) });
                     }
                 }
 
@@ -490,7 +488,7 @@ export class GatewayApiClient {
                     const approveTxHash = await walletClient.writeContract(approveRequest);
                     await publicClient.waitForTransactionReceipt({ hash: approveTxHash, retryCount: RETRY_COUNT });
                 } catch (error) {
-                    throw new ExecuteQuoteError(orderId, translateApprovalError(error));
+                    throw new ExecuteQuoteError(orderId, { cause: translateApprovalError(error) });
                 }
             }
 
@@ -521,7 +519,7 @@ export class GatewayApiClient {
 
                 transactionHash = hash;
             } catch (error) {
-                throw new ExecuteQuoteError(orderId, error);
+                throw new ExecuteQuoteError(orderId, { cause: error });
             }
 
             try {
@@ -626,7 +624,7 @@ export class GatewayApiClient {
                             retryCount: RETRY_COUNT,
                         });
                     } catch (error) {
-                        throw new ExecuteQuoteError(orderId, translateApprovalError(error));
+                        throw new ExecuteQuoteError(orderId, { cause: translateApprovalError(error) });
                     }
                 }
 
@@ -655,7 +653,7 @@ export class GatewayApiClient {
 
                     await publicClient.waitForTransactionReceipt({ hash: txHash, retryCount: RETRY_COUNT });
                 } catch (error) {
-                    throw new ExecuteQuoteError(orderId, translateApprovalError(error));
+                    throw new ExecuteQuoteError(orderId, { cause: translateApprovalError(error) });
                 }
             }
 
@@ -687,7 +685,7 @@ export class GatewayApiClient {
 
                 transactionHash = hash;
             } catch (error) {
-                throw new ExecuteQuoteError(orderId, error);
+                throw new ExecuteQuoteError(orderId, { cause: error });
             }
 
             try {
--- a/sdk/src/gateway/client.ts
+++ b/sdk/src/gateway/client.ts
@@ -75,6 +75,17 @@ function translateApprovalError(error: unknown): unknown {
     return error;
 }
 
+// Extracts `.request` via a generic parameter so the exact overload-resolved type from
+// each simulateContract call site survives across this try/catch, instead of the widened
+// union `Awaited<ReturnType<typeof simulateContract>>` would produce.
+async function simulateApproval<T>(orderId: string, simulate: () => Promise<{ request: T }>): Promise<T> {
+    try {
+        return (await simulate()).request;
+    } catch (error) {
+        throw withOrderId(orderId, translateApprovalError(error));
+    }
+}
+
 /**
  * Resolve the `account` to pass to viem write/simulate/send calls.
  *
@@ -452,15 +463,19 @@ export class GatewayApiClient {
                 // to avoid the ERC20 race condition:
                 // https://github.com/ethereum/EIPs/issues/20#issuecomment-263524729
                 if (needsReset) {
-                    callback?.({ step: 1, type: ExecuteQuoteStepType.ResetApproval, totalSteps, orderId });
-                    try {
-                        const { request: resetRequest } = await publicClient.simulateContract({
+                    const resetRequest = await simulateApproval(orderId, () =>
+                        publicClient.simulateContract({
                             account: signerAccount(walletClient),
                             address: tokenAddress,
                             abi: USDTApproveAbi,
                             functionName: 'approve',
                             args: [spenderAddress, 0n],
-                        });
+                        })
+                    );
+
+                    callback?.({ step: 1, type: ExecuteQuoteStepType.ResetApproval, totalSteps, orderId });
+
+                    try {
                         const resetTxHash = await walletClient.writeContract(resetRequest);
                         await publicClient.waitForTransactionReceipt({ hash: resetTxHash, retryCount: RETRY_COUNT });
                     } catch (error) {
@@ -468,9 +483,8 @@ export class GatewayApiClient {
                     }
                 }
 
-                callback?.({ step: needsReset ? 2 : 1, type: ExecuteQuoteStepType.Approve, totalSteps, orderId });
-                try {
-                    const { request } = await publicClient.simulateContract({
+                const approveRequest = await simulateApproval(orderId, () =>
+                    publicClient.simulateContract({
                         account: signerAccount(walletClient),
                         address: tokenAddress as Address,
                         abi:
@@ -480,8 +494,13 @@ export class GatewayApiClient {
                                 : erc20Abi,
                         functionName: 'approve',
                         args: [spenderAddress, requiredAmount],
-                    });
-                    const approveTxHash = await walletClient.writeContract(request);
+                    })
+                );
+
+                callback?.({ step: needsReset ? 2 : 1, type: ExecuteQuoteStepType.Approve, totalSteps, orderId });
+
+                try {
+                    const approveTxHash = await walletClient.writeContract(approveRequest);
                     await publicClient.waitForTransactionReceipt({ hash: approveTxHash, retryCount: RETRY_COUNT });
                 } catch (error) {
                     throw withOrderId(orderId, translateApprovalError(error));
@@ -606,15 +625,19 @@ export class GatewayApiClient {
                 // to avoid the ERC20 race condition:
                 // https://github.com/ethereum/EIPs/issues/20#issuecomment-263524729
                 if (needsReset) {
-                    callback?.({ step: 1, type: ExecuteQuoteStepType.ResetApproval, totalSteps, orderId });
-                    try {
-                        const { request: resetRequest } = await publicClient.simulateContract({
+                    const resetRequest = await simulateApproval(orderId, () =>
+                        publicClient.simulateContract({
                             account: signerAccount(walletClient),
                             address: tokenAddress as Address,
                             abi: USDTApproveAbi,
                             functionName: 'approve',
                             args: [receiver, 0n],
-                        });
+                        })
+                    );
+
+                    callback?.({ step: 1, type: ExecuteQuoteStepType.ResetApproval, totalSteps, orderId });
+
+                    try {
                         const resetTxHash = await walletClient.writeContract(resetRequest);
                         await publicClient.waitForTransactionReceipt({
                             hash: resetTxHash,
@@ -625,14 +648,8 @@ export class GatewayApiClient {
                     }
                 }
 
-                callback?.({
-                    step: needsReset ? 2 : 1,
-                    type: ExecuteQuoteStepType.Approve,
-                    totalSteps,
-                    orderId,
-                });
-                try {
-                    const { request } = await publicClient.simulateContract({
+                const approveRequest = await simulateApproval(orderId, () =>
+                    publicClient.simulateContract({
                         account: signerAccount(walletClient),
                         address: tokenAddress as Address,
                         abi:
@@ -641,8 +658,18 @@ export class GatewayApiClient {
                                 : erc20Abi,
                         functionName: 'approve',
                         args: [receiver, requiredAmount],
-                    });
-                    const txHash = await walletClient.writeContract(request);
+                    })
+                );
+
+                callback?.({
+                    step: needsReset ? 2 : 1,
+                    type: ExecuteQuoteStepType.Approve,
+                    totalSteps,
+                    orderId,
+                });
+
+                try {
+                    const txHash = await walletClient.writeContract(approveRequest);
 
                     await publicClient.waitForTransactionReceipt({ hash: txHash, retryCount: RETRY_COUNT });
                 } catch (error) {
--- a/sdk/src/gateway/client.ts
+++ b/sdk/src/gateway/client.ts
@@ -316,7 +316,6 @@ export class GatewayApiClient {
 
             let bitcoinTxHex: string;
             if (btcSigner.sendBitcoin) {
-                const sendBitcoin = btcSigner.sendBitcoin.bind(btcSigner);
                 callback?.({
                     step: 1,
                     type: ExecuteQuoteStepType.SignBitcoinTransaction,
@@ -324,7 +323,7 @@ export class GatewayApiClient {
                     orderId,
                 });
                 try {
-                    bitcoinTxHex = await sendBitcoin({
+                    bitcoinTxHex = await btcSigner.sendBitcoin({
                         from: quote.onramp.sender,
                         to: order.onramp.address,
                         value: formatBtc(BigInt(quote.onramp.inputAmount.amount)),
@@ -342,7 +341,6 @@ export class GatewayApiClient {
                         new Error('PSBT not available: sender address is required when using signAllInputs')
                     );
                 }
-                const signAllInputs = btcSigner.signAllInputs.bind(btcSigner);
                 const psbtHex = order.onramp.psbtHex;
                 callback?.({
                     step: 1,
@@ -351,7 +349,7 @@ export class GatewayApiClient {
                     orderId,
                 });
                 try {
-                    bitcoinTxHex = await signAllInputs(psbtHex);
+                    bitcoinTxHex = await btcSigner.signAllInputs(psbtHex);
                     if (!bitcoinTxHex) throw new Error('Failed to get signed transaction');
                 } catch (error) {
                     throw withOrderId(orderId, error);
@@ -509,23 +507,18 @@ export class GatewayApiClient {
             callback?.({ step: totalSteps, type: ExecuteQuoteStepType.SendTransaction, totalSteps, orderId });
             const offrampData = order.offramp.tx.data as Hex;
 
-            let offrampGas: bigint | undefined;
-            if (walletClient.account.type === 'local') {
-                try {
-                    const offrampValue = BigInt(order.offramp.tx.value || 0);
-                    offrampGas = await estimateGasWithBuffer(publicClient, walletClient.account, {
-                        to: spenderAddress,
-                        data: offrampData,
-                        value: offrampValue,
-                    });
-                } catch (error) {
-                    throw withOrderId(orderId, error);
-                }
-            }
-
             let transactionHash: string;
             try {
                 const offrampValue = BigInt(order.offramp.tx.value || 0);
+                const offrampGas =
+                    walletClient.account.type === 'local'
+                        ? await estimateGasWithBuffer(publicClient, walletClient.account, {
+                              to: spenderAddress,
+                              data: offrampData,
+                              value: offrampValue,
+                          })
+                        : undefined;
+
                 const hash = await walletClient.sendTransaction({
                     account: signerAccount(walletClient),
                     data: offrampData,
@@ -680,23 +673,18 @@ export class GatewayApiClient {
             const tokenSwapTo = order.tokenSwap.tx.to as Address;
             const tokenSwapData = order.tokenSwap.tx.data as Hex;
 
-            let tokenSwapGas: bigint | undefined;
-            if (walletClient.account.type === 'local') {
-                try {
-                    const tokenSwapValue = BigInt(order.tokenSwap.tx.value || 0);
-                    tokenSwapGas = await estimateGasWithBuffer(publicClient, walletClient.account, {
-                        to: tokenSwapTo,
-                        data: tokenSwapData,
-                        value: tokenSwapValue,
-                    });
-                } catch (error) {
-                    throw withOrderId(orderId, error);
-                }
-            }
-
             let transactionHash: string;
             try {
                 const tokenSwapValue = BigInt(order.tokenSwap.tx.value || 0);
+                const tokenSwapGas =
+                    walletClient.account.type === 'local'
+                        ? await estimateGasWithBuffer(publicClient, walletClient.account, {
+                              to: tokenSwapTo,
+                              data: tokenSwapData,
+                              value: tokenSwapValue,
+                          })
+                        : undefined;
+
                 const hash = await walletClient.sendTransaction({
                     account: signerAccount(walletClient),
                     data: tokenSwapData,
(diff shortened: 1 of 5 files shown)


danielsimao:
@ducklings-app review

sdk/src/gateway/client.ts
```diff
@@ -46,6 +47,16 @@ import { estimateGasWithBuffer, formatBtc, isValidTronAddress, tronAddressToHex
...
 const RETRY_COUNT = 8; // Number of times to retry fetching transaction receipt after sending a transaction
 
+/**
+ * Tags an error thrown after order creation with the Gateway `orderId`, mutating it in
+ * place so its identity, message, and stack are preserved for existing callers.
+ */
+function attachOrderId(error: unknown, orderId: string): void {
+    if (error instanceof Error) {
+        (error as ExecuteQuoteError).orderId = orderId;
+    }
```
slavastartsev:
```suggestion
    if (error instanceof ExecuteQuoteError) {
        error.orderId = orderId;
    }
```
otherwise runtime deopts from hidden classes chain

sdk/src/gateway/types/execute.ts
```diff
@@ -9,4 +9,19 @@ export interface ExecuteQuoteStep {
...
+    orderId?: string;
+}
+
+/**
+ * Error thrown by {@link GatewayApiClient.executeQuote} once order creation has succeeded.
+ * Carries the Gateway `orderId` so failures during signing, approval, or send can be
+ * cross-referenced against the order record, without altering the original error's
+ * identity, message, or stack.
+ */
+export interface ExecuteQuoteError extends Error {
```
slavastartsev: there should be a class that instantiates an implementation of that interface.

sdk/src/gateway/client.ts
```diff
@@ -333,123 +361,130 @@ export class GatewayApiClient {
...
+                    // Best-effort: the on-chain tx already succeeded, so return the result
+                    // even if registration fails. The order can be reconciled later.
+                }
 
-            return { order, tx: transactionHash };
+                return { order, tx: transactionHash };
+            } catch (error) {
+                attachOrderId(error, orderId);
+                throw error;
+            }
```
slavastartsev: it's allowed in javascript
but design wise you _**should**_ never do it

sdk/src/gateway/client.ts
```diff
@@ -507,96 +542,111 @@ export class GatewayApiClient {
...
```
slavastartsev: ⚠️

sdk/src/gateway/client.ts
```diff
@@ -273,50 +284,67 @@ export class GatewayApiClient {
...
+                if (!instanceOfRegisterTxOneOf(tx)) {
+                    throw new Error('Invalid registerTx response type');
+                }
 
-            return { order, tx: tx.onramp.txid };
+                return { order, tx: tx.onramp.txid };
+            } catch (error) {
+                attachOrderId(error, orderId);
+                throw error;
+            }
```
slavastartsev, line 380: ⚠️

slavastartsev (changes requested):

sdk/src/gateway/client.ts
```diff
@@ -46,6 +47,16 @@ import { estimateGasWithBuffer, formatBtc, isValidTronAddress, tronAddressToHex
 
 const RETRY_COUNT = 8; // Number of times to retry fetching transaction receipt after sending a transaction
 
+/**
+ * Tags an error thrown after order creation with the Gateway `orderId`, mutating it in
+ * place so its identity, message, and stack are preserved for existing callers.
+ */
+function attachOrderId(error: unknown, orderId: string): void {
+    if (error instanceof Error) {
+        (error as ExecuteQuoteError).orderId = orderId;
```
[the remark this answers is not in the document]
danielsimao: Fixed — replaced the mutation with `new ExecuteQuoteError(orderId, error)`, constructed inside a `withOrderId(orderId, operation)` wrapper. Construction can no longer throw regardless of whether the caught value is frozen/non-extensible, since it only reads `error` as the `cause` option (no property assignment on the caught value). See b0c907a.

sdk/src/gateway/client.ts
```diff
@@ -46,6 +47,16 @@ import { estimateGasWithBuffer, formatBtc, isValidTronAddress, tronAddressToHex
 
 const RETRY_COUNT = 8; // Number of times to retry fetching transaction receipt after sending a transaction
 
+/**
+ * Tags an error thrown after order creation with the Gateway `orderId`, mutating it in
```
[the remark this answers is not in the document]
danielsimao: Fixed — `withOrderId` no longer guards on `instanceof Error`. Any caught value (an `Error`, or a plain RPC-rejection object like `{ code, message }` from OKX/Reown wallet adapters) is wrapped in `new ExecuteQuoteError(orderId, error)` with the original preserved via `cause`. Added a test asserting a plain-object rejection from `walletClient.sendTransaction` gets tagged correctly.

sdk/src/gateway/client.ts
danielsimao: `ExecuteQuoteError` is now a real class (see the reply on execute.ts#L25), so `instanceof ExecuteQuoteError` is available. I went further than narrowing this check though: mutation is gone entirely — `withOrderId` always constructs a fresh `new ExecuteQuoteError(orderId, error)` instead of mutating the caught value in place, so this narrowing branch doesn't apply anymore. Original error is preserved via `cause`.

sdk/src/gateway/types/execute.ts
danielsimao: Done — `ExecuteQuoteError` is now a class extending `Error` (sdk/src/gateway/types/execute.ts), exported as a value (not `export type`) from the gateway barrel, so `instanceof ExecuteQuoteError` works at runtime for SDK consumers.

sdk/src/gateway/client.ts
danielsimao: Fixed — dropped the single method-wide try/catch. Post-order-creation, each fallible operation (the signer call/registration call for onramp; approval simulate+write+wait and the send for offramp/tokenSwap) is now wrapped individually via `withOrderId(orderId, () => ...)`, so a failure is tagged at the operation that actually threw rather than one big catch spanning unrelated pre-flight checks and the whole flow body. Same fix applied identically on all three flows (onramp/offramp/tokenSwap).

sdk/src/gateway/client.ts
danielsimao: Same fix as the onramp comment — the outer try/catch is gone; the allowance readContract, the reset+approve sequence, and the send+wait are each wrapped narrowly with `withOrderId`. The best-effort `registerTxV3` catch stays untouched/empty as before.

sdk/src/gateway/client.ts
danielsimao: Same fix as the onramp/offramp comments. The needsApproval block (including the existing ContractFunctionExecutionError → friendlier 'Insufficient native funds' rewrite) is wrapped as one `withOrderId` unit so that rewritten error still ends up tagged with `orderId` via its `cause` chain; the send+wait is a separate narrow `withOrderId` wrap. The best-effort `registerTxV3` catch stays untouched.

danielsimao:
Pushed a follow-up commit (b0c907a) addressing both review passes:

- **`ExecuteQuoteError` is now a real class** (`sdk/src/gateway/types/execute.ts`), not a TS interface, so `instanceof ExecuteQuoteError` works at runtime. It's exported as a value from the gateway barrel.
- **Dropped the per-flow "wrap the whole method body" try/catch.** Each of onramp/offramp/tokenSwap now wraps only the fallible post-order-creation operations narrowly — signer calls, `simulateContract`, `writeContract`, `waitForTransactionReceipt`, and the registration call — via a shared `withOrderId(orderId, operation)` helper.
- **No more mutation.** `withOrderId` constructs `new ExecuteQuoteError(orderId, error)` instead of assigning `.orderId` onto the caught value, so it can't throw a TypeError on frozen/non-extensible errors, and it no longer needs an `instanceof Error` guard — plain-object rejections (e.g. `{ code, message }` from OKX/Reown wallet adapters, the PR's primary motivating case) are now tagged too. The original error is preserved via `cause` rather than folded into `.message`, since a hostile `message` getter could itself throw during construction.
- The existing `ContractFunctionExecutionError` → "Insufficient native funds..." rewrite in the tokenSwap approval step is preserved; it's now inside its own `withOrderId` wrap so the rewritten error still ends up tagged, with the original captured via its `cause` chain.
- The best-effort `registerTxV3` empty-catch blocks are untouched.
- Tests extended: `instanceof ExecuteQuoteError`, `.orderId` correctness, `.cause` identity, a plain-object (non-Error) rejection getting wrapped, and construction from a frozen `Error` not throwing.

Replied inline on each specific finding. `pnpm run build` / `pnpm run lint` / `pnpm run test` all green.

danielsimao:
@ducklings-app review

sdk/src/gateway/client.ts
```diff
@@ -46,6 +47,21 @@ import { estimateGasWithBuffer, formatBtc, isValidTronAddress, tronAddressToHex
 
 const RETRY_COUNT = 8; // Number of times to retry fetching transaction receipt after sending a transaction
 
+/**
+ * Runs a single post-order-creation operation (a signer call, an on-chain write, a
+ * registration call, ...) and rethrows any failure as an {@link ExecuteQuoteError} carrying
+ * the Gateway `orderId`, so it can be cross-referenced against the order record regardless
+ * of whether the underlying rejection was an `Error` (wallet SDKs) or a plain object (some
+ * wallet RPC rejections, e.g. `{ code, message }`).
+ */
```
slavastartsev: 🧹

sdk/src/gateway/types/execute.ts
```diff
@@ -9,4 +9,26 @@ export interface ExecuteQuoteStep {
...
+    orderId?: string;
+}
+
+/**
+ * Error thrown by {@link GatewayApiClient.executeQuote} when a signing, approval, or send
+ * operation fails after order creation has succeeded. Carries the Gateway `orderId` so the
+ * failure can be cross-referenced against the order record. The original error is preserved
+ * as `cause` rather than folded into `message`, since a hostile/custom `message` getter on
+ * the original error could itself throw during construction.
+ */
```
slavastartsev: 🧹

danielsimao:
@ducklings-app review

sdk/src/gateway/client.ts
```diff
@@ -376,40 +426,42 @@ export class GatewayApiClient {
             const totalSteps = needsReset ? 3 : needsApproval ? 2 : 1;
 
             if (needsApproval) {
-                // To change the USDT approval, first set the allowance to 0 (approve(_spender, 0))
-                // to avoid the ERC20 race condition:
-                // https://github.com/ethereum/EIPs/issues/20#issuecomment-263524729
-                if (needsReset) {
-                    const { request: resetRequest } = await publicClient.simulateContract({
+                await withOrderId(orderId, async () => {
```
slavastartsev: please use `withOrderId` only to wrap unhandled error in a catch block

slavastartsev (changes requested):

danielsimao:
@ducklings-app review

Context for this pass: restructured `withOrderId` per prior review feedback — it's now a plain synchronous error-wrapper (`throw withOrderId(orderId, error)`) called individually from a narrow `catch` at each fallible post-order-creation operation in onramp/offramp/tokenSwap, not a closure runner. `ExecuteQuoteError.message` now forwards `cause.message` (defensively, via `getExecuteQuoteErrorMessage`) so translated/user-facing messages (e.g. the tokenSwap insufficient-funds message via the new `translateApprovalError` helper) surface at the top level instead of being buried in `.cause.cause`. Callback invocations are never wrapped. This should address all 3 prior findings (P1 message-burying, P2 callback inconsistency, P2 unwrapped onramp validation throws) plus the human reviewer's "wrap only in a catch block" structural ask.

sdk/src/gateway/types/execute.ts
```diff
@@ -9,4 +9,33 @@ export interface ExecuteQuoteStep {
...
+// A custom/proxy `message` getter on `cause` could throw here, which must never happen
+// during ExecuteQuoteError construction.
+function getExecuteQuoteErrorMessage(cause: unknown): string {
+    try {
+        if (cause instanceof Error && typeof cause.message === 'string') {
+            return cause.message || EXECUTE_QUOTE_ERROR_MESSAGE;
+        }
+    } catch {
+        // fall through
+    }
```
slavastartsev: nothing throws in `try` block
why is it needed?

danielsimao: `cause.message` isn't guaranteed safe to read.

and

ESLint's `no-empty` rule bans `catch {}` with nothing inside, so needs a body

sdk/src/gateway/client.ts
```diff
@@ -58,10 +59,14 @@ function withOrderId(orderId: string, error: unknown): ExecuteQuoteError {
...
-// viem throws ContractFunctionExecutionError for insufficient gas funds too, not just reverts.
-// https://github.com/wevm/viem/blob/3aa882692d2c4af3f5e9cc152099e07cde28e551/src/actions/public/simulateContract.test.ts#L711
+// ContractFunctionExecutionError wraps any contract-call failure (reverts, ABI/decode
+// failures, node errors, insufficient funds, ...), so only translate when InsufficientFundsError
+// is actually in the cause chain — otherwise genuine reverts get mislabeled as a funds issue.
 function translateApprovalError(error: unknown): unknown {
-    if (error instanceof ContractFunctionExecutionError) {
+    if (
+        error instanceof ContractFunctionExecutionError &&
+        error.walk((cause) => cause instanceof InsufficientFundsError)
```
danielsimao, line 61: @slavastartsev what do you think about this? Narrows it down, which I think it is right because `instanceof ContractFunctionExecutionError` matches way more than insufficient funds.

danielsimao:
@ducklings-app review

sdk/src/gateway/types/execute.ts
slavastartsev: should it then use `Object.hasOwn(cause, 'message')` to guard against unsafe access?
the point I'm raising this is because it's not trivial to figure out what `try/catch` is trying to prevent.

on the other hand take a look in MDN docs for Error class https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error#instance_properties
if `cause` is instanceof `Error`, then it's guaranteed to have at least `cause` and `message` properties, so it must be a safe read

sdk/src/gateway/client.ts
slavastartsev: this is perfectly fine 👍

sdk/src/gateway/types/execute.ts
danielsimao: `instanceof Error` only confirms `Error.prototype` is in the prototype chain — it does not guarantee safe property access on `message`. A subclass can define a throwing getter:

```ts
class HostileError extends Error {
    get message(): string { throw new Error("boom"); }
}
new HostileError() instanceof Error; // true — still throws on .message
```

`Object.hasOwn(cause, "message")` does not fix this either — it only confirms an own property exists, not that its a safe data property vs. a throwing accessor. `Object.defineProperty` can turn an existing `Error`s own `message` into a throwing getter after construction, and `Object.hasOwn` would still return `true`:

```ts
const e = new Error("x");
Object.defineProperty(e, "message", { get() { throw new Error("boom"); } });
Object.hasOwn(e, "message"); // true — .message still throws
```

Neither `instanceof Error` nor `Object.hasOwn` establishes anything about *how* the property behaves when read, only that it exists somewhere in the shape. The try/catch is what guarantees `ExecuteQuoteError` construction itself can never replace the original failure with an unrelated error from inspecting its message.

danielsimao:
@ducklings-app review

Fixed the ResetApproval/Approve callback-ordering finding: each site now splits into simulate (fails → no callback, no writeContract) → callback → write+receipt (own try/catch), matching pre-PR ordering while still never wrapping callback throws in ExecuteQuoteError.

sdk/src/gateway/types/execute.ts
danielsimao: Confirmed, `instanceof Error` only proves prototype-chain membership, not safe property access. A subclass can override message with a throwing getter, and `Object.hasOwn` would still report it as present right before it throws. The `try/catch` is the actual guarantee here, not a redundant one.

sdk/src/gateway/client.ts
```diff
@@ -46,6 +49,42 @@ import { estimateGasWithBuffer, formatBtc, isValidTronAddress, tronAddressToHex
...
+    }
+
+    return new ExecuteQuoteError(orderId, error);
+}
+
+// ContractFunctionExecutionError also wraps plain reverts, so only translate when
+// InsufficientFundsError is actually in the cause chain.
+function translateApprovalError(error: unknown): unknown {
+    if (
+        error instanceof ContractFunctionExecutionError &&
```
slavastartsev, line 60: tbh
it would be much better to check if `error instanceof BaseError`

because `walk` is defined in `BaseError`'s prototype

sdk/src/gateway/client.ts
```diff
@@ -46,6 +49,42 @@ import { estimateGasWithBuffer, formatBtc, isValidTronAddress, tronAddressToHex
 
 const RETRY_COUNT = 8; // Number of times to retry fetching transaction receipt after sending a transaction
 
+// Call from the catch block of a single fallible operation, not as a closure runner —
+// keeps each failure attributable to the call that produced it.
+function withOrderId(orderId: string, error: unknown): ExecuteQuoteError {
+    if (error instanceof ExecuteQuoteError && error.orderId === orderId) {
+        return error;
+    }
```
slavastartsev: this `if` is unnecessary, because `withOrderId` is never called with an instance of `ExecuteQuoteError` as the 2nd argument

sdk/src/gateway/types/execute.ts
```diff
@@ -9,4 +9,25 @@ export interface ExecuteQuoteStep {
...
     type: ExecuteQuoteStepType;
     totalSteps: number;
+    /** Absent for callback invocations before order creation. */
+    orderId?: string;
+}
+
+const EXECUTE_QUOTE_ERROR_MESSAGE = 'Failed to execute Gateway quote after order creation';
+
+function getExecuteQuoteErrorMessage(cause: unknown): string {
+    return cause instanceof Error && typeof cause.message === 'string' && cause.message
```
slavastartsev:
```suggestion
    return cause instanceof Error && typeof cause.message === 'string'
```

sdk/src/gateway/client.ts
```diff
@@ -273,47 +312,84 @@ export class GatewayApiClient {
...
             let bitcoinTxHex: string;
             if (btcSigner.sendBitcoin) {
-                callback?.({ step: 1, type: ExecuteQuoteStepType.SignBitcoinTransaction, totalSteps: 1 });
-                bitcoinTxHex = await btcSigner.sendBitcoin({
-                    from: quote.onramp.sender,
-                    to: order.onramp.address,
-                    value: formatBtc(BigInt(quote.onramp.inputAmount.amount)),
-                    opReturn: order.onramp.opReturnData || undefined,
-                    // isSignet: this.isSignet,
+                const sendBitcoin = btcSigner.sendBitcoin.bind(btcSigner);
```
slavastartsev: please, no `bind`
instead use `btcSigner.sendBitcoin` <-- it's absolutely fine

sdk/src/gateway/client.ts
```diff
@@ -273,47 +312,84 @@ export class GatewayApiClient {
...
+                }
             } else if (btcSigner.signAllInputs) {
                 if (!order.onramp.psbtHex) {
-                    throw new Error('PSBT not available: sender address is required when using signAllInputs');
+                    throw withOrderId(
+                        orderId,
+                        new Error('PSBT not available: sender address is required when using signAllInputs')
+                    );
+                }
+                const signAllInputs = btcSigner.signAllInputs.bind(btcSigner);
```
slavastartsev: ⚠️ please no `bind`

sdk/src/gateway/client.ts
```diff
@@ -535,48 +657,60 @@ export class GatewayApiClient {
...
+                    tokenSwapGas = await estimateGasWithBuffer(publicClient, walletClient.account, {
+                        to: tokenSwapTo,
+                        data: tokenSwapData,
+                        value: tokenSwapValue,
+                    });
+                } catch (error) {
+                    throw withOrderId(orderId, error);
+                }
+            }
+
```
slavastartsev: based on previous implementation `estimateGasWithBuffer` can throw but returns fallback value = `undefined`

this is intentionally done to not abruptly cancel `executeQuote` early and use estimated gas value from `rpc-node` 
so I believe `try/catch` is unnecessary there

sdk/src/gateway/types/execute.ts
danielsimao: Removing it would allow empty error messages which right now are going into the fallback

sdk/src/gateway/client.ts
danielsimao: 1.  `ContractFunctionExecutionError extends BaseError`, so `instanceof ContractFunctionExecutionError` already guarantees `.walk` exists

2. Narrowing to it specifically is deliberate: `simulateContract/writeContract` (the only two call sites this feeds) wrap their failures as `ContractFunctionExecutionError`

sdk/src/gateway/client.ts
```diff
@@ -273,47 +302,82 @@ export class GatewayApiClient {
...
                 });
+                try {
+                    bitcoinTxHex = await btcSigner.sendBitcoin({
+                        from: quote.onramp.sender,
+                        to: order.onramp.address,
+                        value: formatBtc(BigInt(quote.onramp.inputAmount.amount)),
+                        opReturn: order.onramp.opReturnData || undefined,
+                        // isSignet: this.isSignet,
+                    });
+                    if (!bitcoinTxHex) throw new Error('Failed to get signed transaction');
```
slavastartsev: this line is probably not needed
`btcSigner.sendBitcoin` will throw an error if it fails to returns `bitcoinTxHex`

sdk/src/gateway/client.ts
```diff
@@ -273,47 +302,82 @@ export class GatewayApiClient {
...
+                } catch (error) {
+                    throw new ExecuteQuoteError(orderId, error);
+                }
             } else if (btcSigner.signAllInputs) {
                 if (!order.onramp.psbtHex) {
-                    throw new Error('PSBT not available: sender address is required when using signAllInputs');
+                    throw new ExecuteQuoteError(
+                        orderId,
+                        new Error('PSBT not available: sender address is required when using signAllInputs')
+                    );
```
slavastartsev, line 332: this looks like a bad design again
Cause should add a contextual value so
`new Error('PSBT not available: sender address is required when using signAllInputs')` isn't a cause to 
`throw new ExecuteQuoteError`

see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause#description

sdk/src/gateway/client.ts
```diff
@@ -273,47 +302,82 @@ export class GatewayApiClient {
...
+                    throw new ExecuteQuoteError(orderId, error);
                 }
-                callback?.({ step: 1, type: ExecuteQuoteStepType.SignBitcoinTransaction, totalSteps: 1 });
-                bitcoinTxHex = await btcSigner.signAllInputs(order.onramp.psbtHex);
             } else {
-                throw new Error('btcSigner must implement either sendBitcoin or signAllInputs method');
+                throw new ExecuteQuoteError(
+                    orderId,
+                    new Error('btcSigner must implement either sendBitcoin or signAllInputs method')
+                );
```
slavastartsev, line 350: same issue as https://github.com/bob-collective/bob/pull/1149/changes#r3829889411

sdk/src/gateway/client.ts
```diff
@@ -356,12 +422,16 @@ export class GatewayApiClient {
...
-                    });
+                    try {
+                        allowance = await publicClient.readContract({
+                            address: tokenAddress as Address,
+                            abi: erc20Abi,
+                            functionName: 'allowance',
+                            args: [accountAddress, spenderAddress],
+                        });
+                    } catch (error) {
+                        throw new ExecuteQuoteError(orderId, error);
```
slavastartsev: but this is correct 👍

sdk/src/gateway/client.ts
```diff
@@ -273,47 +302,82 @@ export class GatewayApiClient {
...
-                return { order, tx };
+                if (typeof response === 'string') {
+                    tx = response;
+                } else if (!instanceOfRegisterTxOneOf(response)) {
+                    throw new Error('Invalid registerTx response type');
+                } else {
+                    tx = response;
+                }
+            } catch (error) {
+                throw new ExecuteQuoteError(orderId, error);
```
slavastartsev: 👍

sdk/src/gateway/client.ts
```diff
@@ -380,56 +450,79 @@ export class GatewayApiClient {
...
+                    to: spenderAddress,
+                    value: offrampValue,
+                    ...(offrampGas !== undefined && { gas: offrampGas }),
+                });
+
+                await publicClient?.waitForTransactionReceipt({ hash, retryCount: RETRY_COUNT });
+
+                transactionHash = hash;
+            } catch (error) {
+                throw new ExecuteQuoteError(orderId, error);
```
slavastartsev: 👍

sdk/src/gateway/client.ts
```diff
@@ -535,48 +640,55 @@ export class GatewayApiClient {
...
-                        // https://github.com/wevm/viem/blob/3aa882692d2c4af3f5e9cc152099e07cde28e551/src/actions/public/simulateContract.test.ts#L711
-                        // throw new error
-                        throw new Error(
-                            'Insufficient native funds for source and destination gas fees, please add more native funds to your account',
-                            { cause: error }
-                        );
-                    }
-
-                    throw error;
+                    throw new ExecuteQuoteError(orderId, translateApprovalError(error));
```
slavastartsev: quite hard to understand but why this error need to be translated(i.e. wrapped in another error)?

slavastartsev (changes requested):

sdk/src/gateway/client.ts
danielsimao: Checked the OKX adapter, the return type is `Promise<string>` but nothing enforces non-empty. Traced `sendBitcoin` -> `EsploraClient.getTransactionHex` -> its internal `getText`:

```ts
private async getText(url: string): Promise<string> {
    const response = await fetch(url);
    if (!response.ok) throw new Error(response.statusText);
    return await response.text();
}
```

Throws on a non-2xx response, but a 200 with an empty body resolves `response.text()` to `""` without throwing -- no guard against that path. This is a real HTTP response edge case on a third-party adapter, not a hypothetical -- keeping the check.

danielsimao:
@ducklings-app review

Context for this pass: addressed two reviewer findings.

1. `ExecuteQuoteError` now takes an options-object second arg (`{ message: string } | { cause: unknown, message?: string }`) instead of a positional `cause` — the two onramp validation failures (missing psbtHex, no signer method) now carry an explicit message with no `cause` at all, instead of smuggling a message in via a synthetic `Error` passed as `cause` with nothing genuine to wrap. Verified `{ cause: undefined }` still sets an own `cause` property (unlike omitting the option), so the constructor only forwards `cause` to `super()` when actually provided.

2. Investigated whether `translateApprovalError` at the 4 write-stage catches (wrapping `walletClient.writeContract`+`waitForTransactionReceipt`) is reachable, since `ContractFunctionExecutionError` is typically associated with `simulateContract`. Confirmed against installed viem source (`writeContract.js`) that `writeContract` internally wraps any `sendTransaction` failure via `getContractError`, which always returns `ContractFunctionExecutionError` — so an insufficient-funds failure at send time (not just simulate time) can reach these catches. Kept all 4 call sites unchanged.

danielsimao:
@ducklings-app review

Fixed the .cause-depth finding: `translateApprovalError` (constructed a friendly Error and passed it as cause, pushing the real viem error to `.cause.cause` -- but only when translation matched, so depth varied by error content, not code path) replaced with `getApprovalErrorMessage`, returning just the friendly string. `ExecuteQuoteError.cause` is now always the exact error the SDK caught, at every throw site; the friendly message goes through the existing `message` option instead. Confirmed the literal originally-proposed fix (unwrap before passing as cause) would have silently dropped the friendly message -- this keeps it while still flattening the depth.

danielsimao:
@ducklings-app review

Added write-stage coverage for the insufficient-funds translation fix from the last commit: a new test where `simulateContract` succeeds and `writeContract` itself rejects with the insufficient-funds error, confirming `getApprovalErrorMessage` and flat `.cause` identity both hold at the write-stage catches too, not just the simulate-stage one. 96 tests passing.

sdk/src/gateway/client.ts
slavastartsev: can you then validate if string contains only valid utf-8 characters please?

sdk/src/gateway/types/execute.ts
```diff
@@ -9,4 +9,30 @@ export interface ExecuteQuoteStep {
...
+
+/** Thrown by {@link GatewayApiClient.executeQuote} for post-order-creation failures. */
+export class ExecuteQuoteError extends Error {
+    readonly orderId: string;
+
+    constructor(orderId: string, options: ExecuteQuoteErrorOptions) {
+        const cause = 'cause' in options ? options.cause : undefined;
+        const message = options.message ?? getExecuteQuoteErrorMessage(cause);
+
+        super(message, 'cause' in options ? { cause } : undefined);
```
slavastartsev: ⚠️ 

`message` should be declared on `cause` object, ExecuteQuoteError should accept new error message
otherwise both `ExecuteQuoteError` and `cause` provide 2 identical error messages which provides no new information

slavastartsev (approved):

slavastartsev:
don't forget to bump version in package.json if this need to be released asap 🙏

PR #1152, opened 2026-08-24
refactor(sdk): simplify quote execution errors

slavastartsev:
## Summary

- align ExecuteQuoteError constructor with Error message/options arguments
- migrate executeQuote callers while preserving messages, causes, and order IDs
- pass empty signer results to Gateway registration and cover behavior

## Test plan

- pnpm run build
- pnpm run lint
- pnpm run format:check
- pnpm run test (93 passed, 11 skipped)

Stacked on #1149.

### DIFF
--- a/sdk/src/gateway/client.ts
+++ b/sdk/src/gateway/client.ts
@@ -51,6 +51,7 @@ const RETRY_COUNT = 8; // Number of times to retry fetching transaction receipt 
 
 const INSUFFICIENT_FUNDS_MESSAGE =
     'Insufficient native funds for source and destination gas fees, please add more native funds to your account';
+const EXECUTE_QUOTE_ERROR_MESSAGE = 'Failed to execute Gateway quote after order creation';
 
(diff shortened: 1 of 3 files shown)


sdk/src/gateway/client.ts
```diff
@@ -339,14 +340,14 @@ export class GatewayApiClient {
                 });
                 try {
                     bitcoinTxHex = await btcSigner.signAllInputs(psbtHex);
-                    if (!bitcoinTxHex) throw new Error('Failed to get signed transaction');
```
danielsimao, line 342: I dont think we should drop this guard as we discussed previously.

danielsimao (changes requested):

sdk/src/gateway/client.ts
slavastartsev: The idea here is that `signAllInputs` should throw an error communicating that signing failed rather than returning empty string or any arbitrary string

danielsimao (approved):

